PR #99, opened 2026-05-15
Migrate to Python MCP server; preserve TypeScript under legacy/typescript

awkoy:
## Summary

- Move the existing TypeScript MCP server into `legacy/typescript/` (history preserved via `git mv`) and add `legacy/typescript/DEPRECATED.md` with the support policy.
- Bring in the Python MCP implementation as the primary server at the repo root (`src/opik_mcp/`, `pyproject.toml`, `uv.lock`, `Makefile`, tests).
- Split CI/CD per stack: legacy TS workflows scoped to `legacy/typescript/**`, new Python CI/release workflows at root (PyPI Trusted Publishing on `py-v*` tags).
- `docs/` is gitignored — internal planning/design notes stay local and will be re-introduced piecemeal in follow-ups.

The TypeScript v2 distribution on npm (`opik-mcp@^2`) continues to work unchanged — main's CI does not run against this branch until merge.

## Test plan

- [ ] Python CI green on this PR (lint + typecheck + tests via `make check`)
- [ ] Legacy TS CI **does not** trigger on this PR (no changes under `legacy/typescript/**`); verify by checking the Checks tab
- [ ] `uvx --from . opik-mcp` runs locally against the new root package
- [ ] `npx -y opik-mcp@2` still resolves to the existing npm release
- [ ] Asset links in `legacy/typescript/README.md` and `legacy/typescript/server.json` resolve after merge (they point at `legacy/typescript/docs/assets/...`)
- [ ] After merge: configure PyPI Trusted Publishing for environments `pypi` / `testpypi`, then dispatch `python-release.yml` with `target=testpypi` as a dry run before tagging `py-v0.1.0`


src/opik_mcp/server.py
```diff
@@ -0,0 +1,494 @@
...
+
+class BearerAuthMiddleware(BaseHTTPMiddleware):
+    def __init__(self, app: ASGIApp, token: str) -> None:
+        super().__init__(app)
+        self._token = token
+
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
+        if request.headers.get("authorization", "") != f"Bearer {self._token}":
+            return JSONResponse({"error": "unauthorized"}, status_code=401)
+        return await call_next(request)
```
JetoPistola, line 524: 🚫 **blocker** | Security

The bearer-token check uses a plain `!=` string comparison, which is variable-time and leaks information about how many leading characters of a guess match the real token. Even on localhost-only Phase 1 this is reachable from any other process on the box, and the same code path will be promoted to a hosted multi-tenant surface in Phase 2 per `README.md`.

Use a constant-time comparison instead:

```suggestion
import secrets


class BearerAuthMiddleware(BaseHTTPMiddleware):
    def __init__(self, app: ASGIApp, token: str) -> None:
        super().__init__(app)
        self._token = token

    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
        auth = request.headers.get("authorization", "")
        expected = f"Bearer {self._token}"
        if not secrets.compare_digest(auth, expected):
            return JSONResponse({"error": "unauthorized"}, status_code=401)
        return await call_next(request)
```

(Don't forget to `import secrets` at module top.)

🤖 *Review posted via /review-github-pr*

src/opik_mcp/server.py
```diff
@@ -0,0 +1,494 @@
...
+
+
+def build_app() -> Starlette:
+    app = mcp.streamable_http_app()
+    app.add_middleware(BearerAuthMiddleware, token=get_settings().opik_mcp_dev_token)
```
JetoPistola, line 532: 💡 **suggestion** | Operability

`build_app()` snapshots `get_settings().opik_mcp_dev_token` into the middleware constructor. Combined with `get_settings()` being `lru_cache`'d, this means the token is fixed for the process lifetime — there is no way to rotate it without a full restart. Probably fine for Phase 1, but worth noting in `docs/` or as a deliberate limitation in `__main__.main` because the warning at startup implies a softer guarantee than this code provides.

No fix required; flagging so it's a conscious choice.

🤖 *Review posted via /review-github-pr*

src/opik_mcp/__main__.py
```diff
@@ -0,0 +1,79 @@
...
+        settings.opik_mcp_port,
+        settings.opik_mcp_reload,
+    )
+
+    if settings.opik_mcp_dev_token == INSECURE_DEFAULT_TOKEN:
+        logger.warning(
+            "OPIK_MCP_DEV_TOKEN is using the insecure default %r. "
+            "Set a strong secret before exposing this server beyond localhost.",
+            INSECURE_DEFAULT_TOKEN,
+        )
```
JetoPistola, line 70: ❓ **question** | Security policy

The HTTP mode logs a warning when `OPIK_MCP_DEV_TOKEN` is the insecure default, but still binds and starts serving. With the default token `"dev-token-123"` baked into source and documented in `README.md` (line 136 and the env-var table), anyone scanning the network for the dev token can authenticate.

The default `OPIK_MCP_HOST=127.0.0.1` mitigates this on a single-user box, but the moment someone exposes it (Docker `-p`, SSH tunnel reverse, `0.0.0.0` override) they're trivially compromised. Should this fail-closed when:

- `opik_mcp_dev_token == INSECURE_DEFAULT_TOKEN` **and**
- `opik_mcp_host not in {"127.0.0.1", "::1", "localhost"}`

? A loud refusal with a remediation message seems safer than a warning that scrolls past in logs.

🤖 *Review posted via /review-github-pr*

.github/dependabot.yml
```diff
@@ -1,10 +1,30 @@
...
-  - package-ecosystem: "npm"
+  # Python (root)
+  - package-ecosystem: "pip"
     directory: "/"
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "python"
```
JetoPistola: 💡 **suggestion** | CI/Dependencies

Dependabot's `pip` ecosystem reads `pyproject.toml` and opens PRs that bump the version constraints there, but it doesn't know how to update `uv.lock`. Since `make check` runs `uv sync --extra dev` (which is strict against the lockfile), every Dependabot PR will fail CI until a human runs `uv lock` locally and pushes the lockfile update.

Options:
1. Add a step in `python-ci.yml` (or a separate workflow on `dependencies` label) that runs `uv lock` and amends the Dependabot PR.
2. Switch to renovate, which has native uv support (https://docs.renovatebot.com/modules/manager/pep621/).
3. Drop the pip ecosystem entry until uv lockfile support lands in Dependabot (https://github.com/dependabot/dependabot-core/issues/10478 — still open as of Jan 2026).

As configured, this will produce 10/week broken PRs.

🤖 *Review posted via /review-github-pr*

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,30 @@
+name: Python CI
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths-ignore:
+      - 'legacy/**'
+      - '.github/workflows/legacy-ts-*.yml'
```
JetoPistola: 💡 **suggestion** | CI / required checks

`paths-ignore` on both `python-ci.yml` and `paths` on `legacy-ts-ci.yml` creates a gap: a PR that only touches top-level files outside both filters (e.g. `README.md`, `CODEOWNERS`, `dependabot.yml`, or the `python-release.yml` workflow itself) will trigger **neither** CI. If `Python CI / ci` is set as a required status check on `main`, those PRs will hang at "Expected — Waiting for status to be reported" indefinitely.

Fix options:
- Drop `paths-ignore` and let Python CI run on every PR (cheapest and most defensive — the lint+typecheck+test cycle is fast).
- Add a no-op job that matches the inverse path set and reports success, so the required check always resolves.
- Use GitHub's required-check-on-skipped-jobs feature with a status-reporter action.

🤖 *Review posted via /review-github-pr*

.github/workflows/python-release.yml
```diff
@@ -0,0 +1,83 @@
+name: Python Release
+
+on:
+  push:
+    tags:
+      - "py-v*"
```
JetoPistola, line 6: 💡 **suggestion** | Release safety

`legacy-ts-deploy.yml` triggers on tags matching `v*` and `python-release.yml` triggers on `py-v*`. The asymmetry is foot-gunny: someone releasing the Python package will reach for `git tag v0.1.0` (the conventional shape, and what historical opik-mcp releases used), which silently fires the **legacy TS deploy**, skips Python entirely, and may publish a stale or duplicate npm version.

Rename the legacy trigger to `npm-v*` (symmetric with `py-v*`) so neither default is the "trap" choice:

```yaml
# legacy-ts-deploy.yml
on:
  push:
    tags:
      - "npm-v*"
```

And update `legacy/typescript/DEPRECATED.md` and any release-runbook docs accordingly. Doing this now is much cheaper than after the first miss-tagged release.

🤖 *Review posted via /review-github-pr*

.github/labeler.yml
```diff
@@ -1,31 +1,39 @@
...
-      - 'src/utils/api.ts'
-      - 'src/tools/**/*'
-      - 'docs/configuration.md'
-      - 'docs/remote-deployment.md'
+      - 'src/opik_mcp/config.py'
+      - 'src/opik_mcp/server.py'
+      - 'src/opik_mcp/comet_client.py'
+      - 'src/opik_mcp/opik_client.py'
+      - 'src/opik_mcp/ollie_client.py'
+      - 'docs/auth-flow.md'
```
JetoPistola: 🧹 **nit** | Config hygiene

Both the `security` label glob (`docs/auth-flow.md`) and the `documentation` label glob (`docs/**/*`) reference paths under `docs/`, but `docs/` is listed in `.gitignore` as "Local-only docs (keep on disk, never tracked)." Those entries will never match a PR's changed files and are dead config.

Either:
- Remove the `docs/...` patterns until docs are re-introduced piecemeal (per PR body).
- Or, if the plan is to bring some docs back in tracked, add a comment here explaining which paths are placeholders for that future state.

🤖 *Review posted via /review-github-pr*

README.md
```diff
@@ -1,236 +1,211 @@
...
+Hosted Model Context Protocol server for **Opik** + **Ollie**. Exposes a curated, OAuth-secured surface to external AI hosts (Claude Code, Cursor, claude.ai, VS Code Copilot, MCP Inspector).
 
-### 2. Add to your MCP client
+| | |
+|---|---|
+| **Status** | Phase 1 in planning — auth-flow research done, PR opened against `comet-backend` |
+| **Jira** | [OPIK-6439](https://www.atlassian.com/) — _Improve Opik MCP server and include Ollie tool_ |
+| **Team brief** | [`docs/team-brief.md`](./docs/team-brief.md) — shareable narrative + team-by-team build list |
+| **Design doc** | [`docs/design.md`](./docs/design.md) — engineering source-of-truth (sequence diagrams, schemas, full task list) |
+| **Notion mirror** | [Notion page](https://www.notion.so/35f7124010a381ca82a5df67d7474313) — same team brief, easier to share |
```
JetoPistola: ❓ **question** | Docs

The PR body says `docs/` is gitignored and "internal planning/design notes stay local and will be re-introduced piecemeal in follow-ups." But `README.md` links to `docs/team-brief.md`, `docs/design.md`, `docs/phase-1.md`, `docs/auth-flow.md`, `docs/decisions/0001-...md`, `docs/decisions/0004-...md`, `docs/decisions/0005-ask-ollie-yolo-mode.md`, `docs/superpowers/plans/2026-05-15-mcp-analytics.md`, and the `docs/install/` placeholder. After merge, all of those will 404 for anyone landing on the repo from PyPI / npm / Notion.

Is the plan to:
1. Strip the README links before merge and add them back as each doc lands?
2. Land a minimal `docs/` skeleton in this PR so the README isn't broken on day one?
3. Accept the broken links as a known interim state?

Flagging because the README will be the public face of the repo as soon as Phase 1 ships to PyPI.

🤖 *Review posted via /review-github-pr*

JetoPistola:
👋 **Review summary**

**What looks good**

- Clean separation between the TS rename and the Python addition — the `git mv` preserves history and the `legacy/typescript/` scoping keeps the old surface untouched, which makes future bisect / blame painless.
- The `audit.py` reasoning is excellent: pinning `_audit_logger.setLevel(INFO)` with a dedicated handler so the log level can't silently drop the audit row under always-on YOLO is the kind of detail that prevents real incidents, and the inline comment explains *why* clearly enough that a future maintainer won't accidentally undo it.
- `ask_ollie.py` handles a genuinely hard set of edge cases (heartbeat vs. real-event monotonic progress counter, `BaseExceptionGroup` unwrap, `confirm_required` dedup-by-`tool_use_id` to keep non-idempotent pod writes from firing twice, separate idle-timeout watchdog from progress heartbeat). The invariants are spelled out in comments where they matter — good defensive style.
- `EntityHandler` registry is the right shape: adding a new readable entity is a one-entry diff, the protocol typing keeps the read/list tools decoupled from the HTTP client, and `id_only` flags save a round-trip on the common case.
- Per-tool analytics extras (`_read_props`, `_score_props`, etc.) bucket aggressively (`bucket_text_len`, `bucket_count`, `canonical` score names) — input prose never leaves the box, which matches the privacy claim in `README.md`.
- Layered config (`Settings` → `resolve_opik_config` → `make_opik_client`) keeps the config-derivation rule in exactly one place; the explicit guard against `COMET_URL_OVERRIDE=""` is a nice catch.

**Overall**

The Python implementation is well-structured for a Phase-1 PoC, with strong observability and audit hygiene that will pay off in the hosted Phase-2 surface. The most material concerns are around the *boundary* — bearer auth and CI/release wiring — rather than the core MCP logic, which is in good shape. None of the inline comments are merge-blocking except the timing-attack fix, which is a small, contained change.

**Inline comments**: 1 blocker, 4 suggestions, 2 nits, 2 questions.

🤖 *Review posted via /review-github-pr*

src/opik_mcp/server.py
awkoy: Fixed in 02efb5233d62e126f239f23402afe80678f81ac0 — `server.py` now uses `secrets.compare_digest(auth, self._expected)` (with `import secrets` at module top). Same shape as your suggestion; verified by the existing `tests/test_http_auth.py` (no-auth → 401, wrong-token → 401, correct-token → 200).

src/opik_mcp/server.py
awkoy: Addressed in 02efb5233d62e126f239f23402afe80678f81ac0 — added a comment in `build_app()` documenting that the bearer token is captured at construction time and rotation requires a server restart (since `get_settings()` is `lru_cache`'d). The middleware now uses `secrets.compare_digest` against a pre-formatted expected header so we don't reach into settings on every request. Phase 2 (OAuth) is the proper fix for live rotation.

src/opik_mcp/__main__.py
awkoy: Fixed in 02efb5233d62e126f239f23402afe80678f81ac0 — `__main__.py` now fails closed when `OPIK_MCP_DEV_TOKEN` is the insecure default *and* `OPIK_MCP_HOST` is not in `{127.0.0.1, ::1, localhost}`. It logs an actionable error and `sys.exit(1)`s. The loopback-only case still warns but starts, since that's the documented dev workflow.

.github/dependabot.yml
awkoy: Addressed in 02efb5233d62e126f239f23402afe80678f81ac0 — commented out the pip ecosystem block in `.github/dependabot.yml` with an inline reference to https://github.com/dependabot/dependabot-core/issues/10478. Dependabot doesn't update `uv.lock` yet, so enabling the ecosystem would only generate noisy PRs against `pyproject.toml` that leave the lockfile stale. We'll re-enable once upstream lands uv support.

.github/workflows/python-ci.yml
awkoy: Fixed in 02efb5233d62e126f239f23402afe80678f81ac0 — removed the `paths-ignore` block from `.github/workflows/python-ci.yml` and left a comment explaining why. Python CI now runs on every PR; the small cost is worth not having a path-filter foot-gun that lets a Python-touching PR sneak through under a docs-only path.

.github/workflows/python-release.yml
awkoy: Fixed in 02efb5233d62e126f239f23402afe80678f81ac0 — `legacy-ts-deploy.yml` trigger renamed `v*` → `npm-v*` and the three `${GITHUB_REF#refs/tags/v}` substrings updated to strip the `npm-v` prefix. `legacy/typescript/DEPRECATED.md` now documents the convention: legacy npm releases use `npm-v*`, future Python releases will use `py-v*`. No more overlap with the historical `v*` tags.

.github/labeler.yml
awkoy: Fixed in 02efb5233d62e126f239f23402afe80678f81ac0 — removed `docs/auth-flow.md` from the `security` label and `docs/**/*` from the `documentation` label in `.github/labeler.yml`. `docs/` is gitignored (Notion is our source of truth), so those globs would never match and were misleading.

README.md
awkoy: Fixed in 02efb5233d62e126f239f23402afe80678f81ac0 — `README.md` no longer links to gitignored `docs/*.md` files. The "Repo layout (planned)" section was replaced with the actual current layout, and a note explains that internal design docs live in our Notion mirror. No more broken links from a fresh clone.

alexkuzmik (approved):

PR #139, opened 2026-05-31
[OPIK-6742] [BE] feat: opik-mcp forwards OAuth bearer + serves RFC 9728 protected-resource metadata

LifeXplorer:
## Summary

Implements opik-mcp's resource-server-transport role for the hosted OAuth flow ([OPIK-6742](https://comet-ml.atlassian.net/browse/OPIK-6742)). opik-mcp becomes a thin transport adapter on HTTP transport — no local bearer validation, no permission knowledge, no separate validator round-trip. opik-backend's `AuthFilter` (shipped in [opik#6900](https://github.com/comet-ml/opik/pull/6900)) is the single point of truth and accepts both API keys and `opik_at_…` OAuth tokens, so opik-mcp just forwards.

Companion design: [OPIK-6666 design doc](https://www.notion.so/cometml/Opik-MCP-OAuth-36c7124010a38086b246f41c201ef569).

## What changed

**Inbound auth (`server.py`, new `auth_context.py`)**
- `BearerAuthMiddleware` performs **no local credential validation**: any well-formed `Bearer …` is accepted, captured into ContextVars (`inbound_authorization`, `inbound_workspace`), and forwarded verbatim on the outbound opik-backend call. The backend is the single point of auth enforcement.
- Missing/empty `Authorization` or a non-Bearer scheme → 401 + `WWW-Authenticate: Bearer realm="opik-mcp", resource_metadata="<discovery-url>"` so MCP hosts can bootstrap the OAuth dance per RFC 6750.
- Health probes, the discovery endpoint, and AS-probe paths are auth-exempt.

**Dev-token mode removed (`config.py`, `__main__.py`, `Makefile`)**
- `OPIK_MCP_DEV_TOKEN` is deleted entirely. Its only use case — local HTTP testing — is covered by passthrough: send `Authorization: Bearer <OPIK_API_KEY>` and the backend validates it (comet-backend strips the `Bearer ` prefix; OSS installs don't authenticate at all, so a token gate there was never a real boundary). No deployment manifest anywhere sets it.
- The "refuse to start with default token on non-loopback bind" guard goes with it — passthrough carries no server-side secret, so the security posture is exactly that of the backend behind it. Documented in the README (see below); default bind stays `127.0.0.1`.

**Discovery (`server.py`)**
- New `GET /.well-known/oauth-protected-resource` per RFC 9728. Returns `{authorization_servers: [OPIK_MCP_AS_URL], resource: OPIK_MCP_RESOURCE_URI}`. 503 when `OPIK_MCP_AS_URL` isn't set — makes the misconfiguration loud rather than silently breaking the OAuth bootstrap.
- AS-probe proxying: MCP host SDKs probe `/register`, `/authorize`, `/token`, `/.well-known/oauth-authorization-server`, OIDC fallback, etc. at the RS host pre-token. These are proxied to `OPIK_MCP_AS_URL` (proxy, not redirect — some SDKs refuse cross-origin OAuth-discovery redirects) so split-host deploys work like the production single-edge deploy.
- Starlette's plain-text 404 replaced with a JSON 404 — SDKs that JSON-parse every discovery response choke on `text/plain`.

**Bearer forwarding (`opik_client.py`)**
- `resolve_opik_config()` prefers the inbound `Authorization` + `Comet-Workspace` headers (via ContextVars) over the env-bound `OPIK_API_KEY` / `COMET_WORKSPACE`. Same code path handles both API keys and `opik_at_…` tokens — opik-backend's `AuthFilter` already discriminates server-side.
- For `opik_at_…` tokens the workspace is optional client-side (the backend derives it from the token row); `Comet-Workspace` is omitted from outbound headers when absent. `run_experiment` still requires a workspace (needed client-side for the summary URL) and fails with a clear error when neither env nor header provides one.

**New config (`config.py`)**
- `OPIK_MCP_AS_URL: str | None = None` — advertised as `authorization_servers` in the discovery doc; proxy target for AS probes.
- `OPIK_MCP_RESOURCE_URI: str | None = None` — advertised as `resource`; derives the `WWW-Authenticate` metadata URL.

**README**
- New "Choosing a transport" section: stdio recommended for same-machine local OSS; stdio + API key or HTTP + OAuth for local→remote; HTTP for hosted opik-mcp behind the same edge as opik-backend. Notes explicitly that OSS backends don't authenticate, so an HTTP opik-mcp in front of one is as open as the OSS REST API itself.
- Env-var table: `OPIK_MCP_DEV_TOKEN` row replaced with `OPIK_MCP_AS_URL` / `OPIK_MCP_RESOURCE_URI`.

## What's NOT in this PR (by design)

- **OAuth on stdio transport.** Out of scope — would require opik-mcp to act as the OAuth client (PKCE dance, browser, OS-keychain token storage). stdio + API key continues to work exactly as before; stdio has no inbound auth at all (whoever can spawn the process owns its stdio).
- **`comet_client.py` bearer forwarding.** `ask_ollie` / `run_experiment` are Comet-Cloud-only and use the Comet API key directly; not affected by this change.
- **Telemetry `auth_mode` tagging.** Will follow in a separate commit.

## Test plan

- [x] `make check` — ruff clean, mypy clean (101 files), 967 passed / 2 skipped
- [x] New tests:
  - `test_oauth_protected_resource.py` (4) — configured/unconfigured/auth-exempt
  - `test_oauth_passthrough_mode.py` (7) — Bearer accepted + ContextVar capture/reset, Comet-Workspace captured, missing-auth → 401 + WWW-Authenticate, malformed-auth rejected, health + discovery paths exempt
  - `test_oauth_redirect.py` (8) — AS-probe proxying: metadata/OIDC/DCR/authorize round-trips, query-string preservation, 503 unconfigured, auth-exempt, JSON 404
  - `test_resource_metadata_url.py` (6) — WWW-Authenticate URL derivation from the resource URI
- [x] `test_http_auth.py` reworked for the single passthrough mode: no-auth → 401, non-Bearer scheme → 401, any Bearer → initialize 200
- [ ] Manual smoke test: opik-mcp HTTP transport against a local OSS opik-backend with `MCP_OAUTH_ENABLED=true` — pending PR landing so we can pick it up in the OPIK-6666 testing matrix

[OPIK-6742]: https://comet-ml.atlassian.net/browse/OPIK-6742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,10 +74,10 @@ docker-build:
 
 docker-run:
 	# Explicit 127.0.0.1 binding: on Linux, `-p 8080:8080` listens on
-	# 0.0.0.0, which combined with the dev-token-123 default token would
-	# expose MCP on every network interface of a dev VM or CI runner.
+	# 0.0.0.0, exposing MCP on every network interface of a dev VM or CI
+	# runner. opik-mcp does no local token validation (the backend does),
+	# so keep the port loopback-only.
 	docker run --rm -p 127.0.0.1:8080:8080 \
-	  -e OPIK_MCP_DEV_TOKEN=$${OPIK_MCP_DEV_TOKEN:-dev-token-123} \
 	  -e COMET_URL_OVERRIDE=$${COMET_URL_OVERRIDE:-https://www.comet.com} \
 	  --name opik-mcp opik-mcp:dev
 
--- a/README.md
+++ b/README.md
@@ -314,8 +314,26 @@ Every setting is an environment variable. Required ones in **bold**.
 | `OPIK_MCP_HOST` | `127.0.0.1` | uvicorn bind host (`streamable-http` only). |
 | `OPIK_MCP_PORT` | `8080` | uvicorn bind port (`streamable-http` only). |
 | `OPIK_MCP_RELOAD` | `false` | `true` to enable uvicorn `--reload` (dev only). |
-| `OPIK_MCP_DEV_TOKEN` | `dev-token-123` | Bearer token the HTTP transport requires. |
+| `OPIK_MCP_AS_URL` | _unset_ | OAuth Authorization Server URL, advertised in `/.well-known/oauth-protected-resource` (RFC 9728) and used as the proxy target for AS-discovery probes. Required for MCP hosts to bootstrap the OAuth dance over HTTP. |
+| `OPIK_MCP_RESOURCE_URI` | _unset_ | Canonical public URI of this server, advertised as `resource` in the protected-resource metadata and used to derive the `WWW-Authenticate` hint. |
 | `OPIK_MCP_LOG_LEVEL` | `INFO` | stderr logger threshold. |
+
+#### Choosing a transport
+
+opik-mcp performs **no local credential validation** on HTTP transport: any
+well-formed `Authorization: Bearer …` (an Opik API key or an `opik_at_…`
+OAuth access token) is forwarded verbatim to opik-backend, which is the
+single point of auth enforcement. Pick the transport by deployment shape:
+
+| Scenario | Transport |
+|---|---|
+| MCP client and Opik on the same machine (local OSS install) | **stdio** (recommended — simplest, no port, no OAuth setup) |
+| Local MCP client → remote Opik (Comet cloud / self-hosted) | stdio with `OPIK_API_KEY`, or HTTP with OAuth (`OPIK_MCP_AS_URL` pointing at the backend) |
+| Hosted opik-mcp behind the same edge as opik-backend | **HTTP** — bearers are validated by the backend per request |
+
+Note for local OSS installs: the OSS backend does not authenticate requests,
+so an HTTP opik-mcp in front of it is as open as the OSS REST API itself.
+Keep the default `127.0.0.1` bind (and prefer stdio) on shared networks.
 
 ### Ollie / long calls
 
--- a/src/opik_mcp/__main__.py
+++ b/src/opik_mcp/__main__.py
@@ -338,6 +338,24 @@ def _run_transport(settings: Settings, transport: str) -> None:
         settings.opik_mcp_reload,
     )
 
+    # OAuth on HTTP transport needs an explicit resource URI. RFC 9728 makes
+    # `resource` REQUIRED in the protected-resource doc, and the AS validates
+    # the authorize `resource` param by exact-equality against its own
+    # MCP_OAUTH_RESOURCE_URI — so a missing value yields a non-compliant doc and
+    # a host-derived fallback that fails every authorize with invalid_target.
+    # Fail fast. (No AS configured = API-key-only mode; resource URI N/A.)
+    if settings.opik_mcp_as_url and not settings.opik_mcp_resource_uri:
+        logger.error(
+            "Refusing to start: OPIK_MCP_AS_URL=%r enables OAuth but "
+            "OPIK_MCP_RESOURCE_URI is unset. Set it to the exact resource URI the "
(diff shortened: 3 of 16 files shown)


awkoy:
Went deep on the MCP OAuth work (this PR + the `OPIK-6666` AS branch) against the latest MCP auth spec ([2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization)) + RFC 9728/8707. Overall it's **really solid** — PKCE S256 enforced, `redirect_uri` validated before trust, CSRF on consent, tokens hashed at rest, refresh rotation w/ reuse-detection, `resource` pinned at `/authorize`, and OAuth tokens still go through `@RequiredPermissions`. 👏

A few things worth a look — fix if you agree:

**1. Re-auth on token expiry looks broken at the opik-mcp edge.** `opik_at_` tokens expire, but opik-mcp forwards them verbatim, so a backend 401 surfaces as a tool/JSON-RPC error *without* `WWW-Authenticate: …resource_metadata=…`. Hosts can bootstrap OAuth on first connect but can't auto-refresh mid-session. opik-mcp should translate an upstream 401 → its own 401 + `WWW-Authenticate`. (spec: *"invalid or expired tokens MUST receive a 401"*)

**2. Confirm `/opik/auth-by-username` is internal-only.** `authorizeOAuth` trusts the `X-Opik-OAuth-Username` header to impersonate the user — if react-service accepts that endpoint/header from anywhere external, it's an auth bypass. Just want to confirm it's network-restricted to opik-backend.

**3. PRM `resource` field + exact-match coupling.** RFC 9728 makes `resource` REQUIRED in the protected-resource doc, but opik-mcp omits it when `OPIK_MCP_RESOURCE_URI` is unset and derives it as scheme+authority (drops the path). The AS does an exact `equals` on `config.getMcpResourceUri()` at `/authorize` → if the advertised resource ≠ the AS value (e.g. AS has a `/api/v1/mcp` path), every authorize fails with `invalid_target`. Suggest making `OPIK_MCP_RESOURCE_URI` required and byte-matching the AS config.

_Minor / optional:_
- RS `validateAccessToken` doesn't re-check `resource` per request — safe today thanks to single-resource issuance, but a cheap audience check would future-proof multi-resource.
- Quick sanity-check the auth code is single-use (consumed on exchange) to block replay.

Happy to pair on any of these 🙌

awkoy (approved):

LifeXplorer:
> **Re-auth on token expiry looks broken at the opik-mcp edge.** `opik_at_` tokens expire, but opik-mcp forwards them verbatim, so a backend 401 surfaces as a tool/JSON-RPC error _without_ `WWW-Authenticate: …resource_metadata=…`. Hosts can bootstrap OAuth on first connect but can't auto-refresh mid-session. opik-mcp should translate an upstream 401 → its own 401 + `WWW-Authenticate`. (spec: _"invalid or expired tokens MUST receive a 401"_)

These are good points raised!
For 
p1: technically correct. But most of popular AI hosts (like Claude, Cursor, etc.) implement MCP OAuth and refresh proactively before expiry. In normal operation they present a fresh token on every call, so the resource server essentially never sees an expired token. I've created a follow-up ticket in case we ever face this issue from the users.
https://comet-ml.atlassian.net/browse/OPIK-6782
p2: Good catch. There is already a ticket for hiding the endpoint: https://comet-ml.atlassian.net/browse/OPIK-6736
p3: Updated

PR #142, opened 2026-06-03
Dockerfile: numeric USER (1000) for runAsNonRoot compatibility

liyaka:
The image set `USER app` (a name). Under Kubernetes `runAsNonRoot: true`, the kubelet verifies non-root from the image config **before** starting the container and can't resolve a name to a UID, so the pod fails:

```
container has runAsNonRoot and image has non-numeric user (app), cannot verify user is non-root
```

Fix: create the user with an explicit `--uid 1000` and set a **numeric** `USER 1000`. The named account still exists (file ownership/`chown app:app` resolves), but the directive is numeric so the check passes — no per-deployment `runAsUser` patching needed.

Verified locally: image builds, runs as `uid=1000(app)`, app imports and version resolves.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ FROM python:3.13-slim-bookworm AS runtime
 RUN apt-get update \
     && apt-get install -y --no-install-recommends tini \
     && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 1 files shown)


liyaka:
Applied directly to main (98e47fd).

PR #148, opened 2026-06-08
feat(analytics): BI events for the hosted auth/HTTPS flow (GAP#1/#2/#3)

awkoy:
## Summary

Extends opik-mcp's BI/analytics so the **hosted HTTP/OAuth deployment** is no longer dark and every event carries the metadata needed for the most useful usage charts. Closes three gaps in the analytics layer relative to the recently-added OAuth resource-server + transport-security work.

**GAP#1 — lifecycle events were dark in hosted mode.** `server_started`/`server_shutdown`/`startup_error` were only emitted from `__main__.main()`, but the Docker/Helm entrypoint runs `uvicorn opik_mcp.server:build_app --factory`, which bypasses `main()`. `build_app()` now composes an analytics lifespan around FastMCP's session-manager lifespan and emits them itself — guarded by an env-var sentinel so a `main()`-driven HTTP boot is never double-counted (verified inherited by `--reload` spawn workers).

**GAP#2 — per-request OAuth identity never reached BI.** `_build_event()` runs synchronously in the request task, so the `auth_context` ContextVars are live: events now carry per-request `auth_mode`, `token_sha256` (hash of the `opik_at_` bearer — raw token never stored), `request_workspace`, and `installation_type`. Extraction mirrors `resolve_opik_config` exactly so BI agrees with the outbound credential.

**GAP#3 — auth failures were invisible.** New `opik_mcp_auth_rejected` event from a pure-ASGI `AuthRejectionMiddleware` (never buffers SSE) for 401/421/403 on authenticated paths, with bucketed `rejection_reason`, `auth_mode`, `path_bucket`, `oauth_configured`, `resource_metadata_url_present`. OAuth-proxy/discovery paths are skipped (a proxied-AS 401 isn't our rejection).

Also fixes a **pre-existing privacy-contract bug**: the `McpHost`/`HostLlmFamily` `Literal` allowlists were out of sync with the classifiers (`codex`/`gemini-cli`/`openai`/`google` were shipping undeclared).

### Per-event additions
- `server_started`: `installation_type`, `oauth_configured`, `resource_uri_scheme`, `dns_rebinding_protection`, `allowed_hosts_is_default`, `auth_mode`, `lifecycle_source`
- `server_shutdown`: `lifecycle_source`; `startup_error`: `oauth_configured` (+ `installation_type` via common)
- `tool_called` / `session_initialized` / `ask_ollie_completed`: `auth_mode`, `token_sha256`, `request_workspace`, `installation_type`
- new `opik_mcp_auth_rejected` event

### Notes
- New `boot_props` module is the single source of truth shared by `main()` and the lifespan so the two boot paths cannot drift. The pure `installation_type` classifier moved to `config` (leaf) to break an import cycle.
- **Cross-team dependency:** `token_sha256` is only a useful BI join key once the backend exposes an `opik_at_ → user` mapping; until then dashboards should treat `auth_mode='oauth'` rows as `identity_pending`.
- Privacy contract upheld: every new property is a bool-string / allowlisted `Literal` / bucket, or a SHA-256 identity hash; raw tokens, paths, and user prose never leave the process. Enforced by direct `_build_event` tests + the cross-event canary sweep.
- Design/plan doc lives at `docs/bi-events-https-plan.md` (gitignored, local).

## Test Plan
- [x] `uv run pytest` — 1061 passed, 2 skipped (clean env)
- [x] `uv run ruff check .` + `ruff format --check .` — clean
- [x] `uv run mypy` — clean (107 files)
- [x] Each change developed test-first (red → green) and code-reviewed; findings fixed
- [ ] Staging: confirm `server_started` (`lifecycle_source=lifespan`) fires within minutes of a Docker `--factory` boot, and `auth_rejected` fires on an unauthenticated request
- [ ] Backend: confirm the `opik_at_ → user` mapping is BI-queryable before relying on `token_sha256`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,11 +67,13 @@ ENV OPIK_MCP_TRANSPORT=http \
 
 STOPSIGNAL SIGTERM
 
-# Single worker by design: SSE streams are async-IO-bound, and multiple
-# workers would fragment in-memory MCP session state across processes.
-# Scale with replicas + Redis (see docs/phase-2.md), not workers.
-ENTRYPOINT ["tini", "--", "python", "-m", "uvicorn", \
-            "opik_mcp.server:build_app", "--factory", \
-            "--host", "0.0.0.0", "--port", "8080", \
-            "--workers", "1", \
-            "--no-access-log"]
+# Run via the package entrypoint (main()) rather than uvicorn directly, so the
+# full BI lifecycle fires in hosted mode exactly as in stdio: main() emits
+# server_started / server_shutdown / startup_error, runs the preflight bind
+# check and the OAuth-config guard, then serves HTTP via uvicorn (single worker,
+# access logging off). The build_app() lifespan defers to main() via the
+# _OPIK_MCP_LIFECYCLE_OWNED_BY_MAIN sentinel, so boots are never double-counted.
+# Transport/host/port come from the ENV above; single worker by design (SSE is
+# async-IO-bound and multiple workers would fragment in-memory session state —
+# scale with replicas + Redis, see docs/phase-2.md).
+ENTRYPOINT ["tini", "--", "python", "-m", "opik_mcp"]
--- a/src/opik_mcp/__main__.py
+++ b/src/opik_mcp/__main__.py
@@ -389,7 +389,15 @@ def _run_transport(settings: Settings, transport: str) -> None:
             reload_dirs=["src"],
         )
     else:
-        uvicorn.run(build_app(), host=settings.opik_mcp_host, port=settings.opik_mcp_port)
+        # access_log=False: parity with the hosted image's old --no-access-log,
+        # and it keeps OAuth-flow query strings (which can carry tokens) out of
+        # stdout. BI + Sentry are the observability surface, not uvicorn logs.
+        uvicorn.run(
+            build_app(),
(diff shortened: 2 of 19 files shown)


awkoy:
Follow-up (commit `d6bc779`): made the **hosted entrypoint run `main()`** (`python -m opik_mcp` instead of `uvicorn … build_app --factory`).

Reason: `opik_mcp_startup_error` (plus the preflight bind check and the OAuth-config guard) is emitted only from `main()`, so under the old `--factory` entrypoint it never fired in hosted mode — the one event the `build_app()` lifespan doesn't cover. Now **every BI event fires identically for stdio (local) and HTTP (hosted)**.

- The `_OPIK_MCP_LIFECYCLE_OWNED_BY_MAIN` sentinel (set before `build_app()`) makes the lifespan defer to `main()` → no double-count (new test asserts the ordering + `lifecycle_source='main'`).
- Parity preserved: single worker (uvicorn default), `access_log=False` (replaces `--no-access-log`; also keeps OAuth-flow query strings out of stdout), host/port/transport from ENV.
- Helm inherits the image entrypoint — no chart change.
- tini stays PID 1 for SIGTERM forwarding → graceful shutdown → `server_shutdown(reason=clean_exit)`.

PR #156, opened 2026-07-30
[OPIK-7473] Build the /instrument skill: configure, instrument, run, verify a real trace

aswynz:
Ships `/instrument` — a skill that takes an existing app from no observability to a first Opik trace you can open, in one command.

## What it does

Run `/instrument` (optionally `/instrument <path>`) and it:

1. **Configures Opik** from whatever source of truth already exists — `~/.opik.config`, `OPIK_API_KEY`, or the official `opik configure` / `npx opik-ts configure` flow. Never introduces a second config mechanism and never copies secrets between them.
2. **Detects the language and LLM framework** — Python or TypeScript, then the framework from its imports (OpenAI, Anthropic, LangChain/LangGraph, CrewAI, DSPy, LlamaIndex, LiteLLM, Bedrock, Vercel AI SDK, and the rest).
3. **Adds the minimum tracing** — the framework-native integration for provider calls, manual `@opik.track` spans only for the orchestration and tools the integration doesn't cover, one entrypoint per independently-runnable agent, and a flush for scripts.
4. **Installs the Opik package** through the repo's own package manager (pip / uv / poetry / npm / pnpm / yarn) and reports it as a change.
5. **Runs a safe representative path** — an existing test, example, or dev script, or a bounded single-request entrypoint.
6. **Verifies the trace landed** by querying Opik with the SDK it just installed, reconstructing the span tree rather than assuming ingestion worked.
7. **Reports** a short human summary with the trace link.

Already-instrumented projects get an audit instead: add what's missing, don't re-instrument.

## What "done" means

A trace confirmed in Opik and a direct link back. If that can't be reached safely or autonomously, the skill stops at the **first** genuine blocker and returns **exactly one** next step — while still reporting the changes it made. Code edits on their own are not success.

## Design choices worth knowing

- **Everything is inferred.** The path, project name, and run command are optional overrides, not setup questions. Inference never becomes a questionnaire.
- **Native integration first, and never the same operation twice.** No manual LLM span stacked on top of `track_openai`.
- **Safety over a trace.** It will not run anything that looks like production or does irreversible or expensive work just to force a trace — that's a blocker instead.
- **Bounded installs.** Only the Opik package, via the detected package manager, lockfile preserved, no global installs and no broad upgrades.
- **Activation only.** No prompt migration, threading, or span-broadening during the run. Once the trace lands, the skill makes a single consolidated offer to go deeper.
- **Human summary on top, small state model underneath** (`verified` / `blocked` / `already_verified` / `unsupported`) so other tooling can compose on it. `blocked` still reports what changed; `unsupported` modifies nothing.

This also retires `skills/opik/instrument-tracing.md`, which the skill supersedes.

## Related

Evals for the skill live in [comet-ml/opik-skill-evals](https://github.com/comet-ml/opik-skill-evals), so the published pack stays skill-only.

Part of "Opik for Agents" (OPIK-7623) · OPIK-7473


awkoy:
Hey @aswynz, heads up on something that landed while this was open.

#163 is merged, so `opik-mcp` now generates the public `opik-skills` pack and CI validates every skill against the official agentskills.io validator. This PR will go red on that check. I ran the validator against your branch to be sure:

```
Unexpected fields in frontmatter: argument-hint, last_updated, source_commit.
Only ['allowed-tools', 'compatibility', 'description', 'license', 'metadata', 'name'] are allowed.
```

The fix is moving those three under `metadata:` with the values unchanged:

```yaml
metadata:
  last_updated: "2026-08-05"
  source_commit: "TODO — pin to the Opik release this was verified against (OPIK-7471)"
  argument-hint: "[optional: file or directory path]"
```

I tried exactly that on a local merge of your branch with main and it goes green, 10 tests pass and the pack builds with all three skills. Same change we already applied to `/opik` and `/evaluate` in #163, so it is consistent rather than a special case.

One thing worth your call. `argument-hint` is a Claude Code field rather than a spec one, so under `metadata:` it stays legal, but I did not check whether Claude Code still picks it up from there. If the slash command hint matters to you, worth confirming before assuming it survives.

Separately, good news on the cross skill references you added in 5792414f. The pack keeps skills as siblings, so `../opik/references/*.md` resolves after install. I checked with the real installer using the same `-g --all` that the onboarding runs, and all four paths resolve. Installing `--skill instrument` on its own does leave them dangling, which your `compatibility` note and the docs fallback already cover.

Last thing so you have the full picture. This PR is now what gates the cutover. `/instrument` exists in `opik-skills` but not yet in `opik-mcp`, so running the sync before this merges would delete it from the public pack. That is why the sync workflow shipped with its schedule commented out. Once this lands, the first sync publishes all three skills together.

aswynz:
Thanks @awkoy — fixed in `64e905c`. Moved `argument-hint`, `last_updated`, and `source_commit` under `metadata:` (values unchanged); `compatibility` and `allowed-tools` stay top-level since they're in the allowed set. Verified the top-level keys now conform, consistent with the `/opik` + `/evaluate` fix in #163.

On the `argument-hint` caveat you flagged — good call, I ran it down against the Claude Code docs. `argument-hint` is a **top-level-only** field; under `metadata:` it is **inert** (CC treats `metadata` as free-form and does not act on it). So this move disables the autocomplete hint in Claude Code. It has no effect on behavior or argument parsing, but it is a real (small) DX affordance — the point-of-invocation discoverability that a path can be passed — so this is a minor DX regression for CC users, not a no-op. The optional path is still documented in the skill body's `## Inputs` section, which covers the agent and anyone reading the file, just not the human at the autocomplete prompt.

I've kept it under `metadata:` here as intent-preserving documentation so the fix stays consistent with `/opik` and `/evaluate` and unblocks the cutover. But it surfaces a broader question that applies to **every** skill in the pack, not just this one: the validator's allow-list omits `argument-hint` (and any other Claude-Code-specific fields), so all skills lose their CC arg hints. Should the spec-compliance allow-list permit known CC fields, or should the `opik-claude-code-plugin` distribution re-inject them at build time so the public pack stays strictly compliant? Raised as **OPIK-7979** to decide and apply that fleet-wide; this PR just takes the interim `metadata:` placement.

Also confirmed the two FYIs: `build_skills_pack.py` already excludes `evals/` (so the OPIK-7800 harness won't bloat the public pack), and the reference check covers the `../opik/references/*.md` paths. Happy to be the gating merge for the cutover.

src/opik_mcp/skills/instrument/SKILL.md
```diff
@@ -0,0 +1,133 @@
...
+The entry point is just `/instrument` (optionally `/instrument <path>`). Infer everything else; treat these only as **optional overrides** the user may pass, never as required setup:
+
+- target path (default: project root) · project name (default: inferred from the repo) · run command (default: an inferred safe path) · `migrate_prompts` (default: **false**).
+
+Never turn inference into a questionnaire. Ask only when you hit a genuine, non-inferable blocker (see **Blockers**).
+
+## Activation — the only in-scope work
+
+### 1. Configure Opik (one source of truth)
+- If `~/.opik.config` exists or `OPIK_API_KEY` is set, use it as-is.
```
alexkuzmik, line 35: Let's add `opik healthcheck` CLI call to verify the config works

alexkuzmik (approved):
looks good, left just one improvement suggestion that should improve configuration step

src/opik_mcp/skills/instrument/SKILL.md
aswynz: Done in de46985 — added `opik healthcheck` to the Configure step, right after `opik configure`. It validates the config, the install, and backend/workspace reachability, and a failure becomes the first Blocker (fix config/connectivity before any tracing). Verified the command exists and does exactly this. Thanks for the review!

PR #167, opened 2026-08-24
[OPIK-7649] Build the /opik-explain skill: root-cause a trace via the SDK, MCP-optional

aswynz:
## What

Adds the **`/opik-explain`** task skill — the second stage of the debug flow. Given a trace id (or a behavior/pattern), the coding agent fetches the trace + spans **via the SDK**, root-causes it against the repo, and returns **root cause + evidence spans + one next step**.

Closes OPIK-7649.

## Design (per the ticket)

- **SDK-first, MCP-optional.** The agent runs the investigation itself over SDK-fetched data; `ask_ollie` is an *optional alternative reasoner* when the hosted MCP is connected, never a gate. Same output contract either way.
- **Read-only by design** — `allowed-tools` is `Read, Grep, Glob, Bash` (no `Edit`/`Write`). `/opik-explain` diagnoses; `instrument`/`test` or the developer make changes.
- **SDK footguns encoded** (learned from the `/instrument` work): `get_trace_content` exposes `project_id`, not `project_name`; `search_spans` is a separate call; reconstruct the tree via `parent_span_id`.
- **One output contract**: `status / root_cause / evidence / next_step / reasoner`.

## Acceptance criteria

- [x] Given a trace id, returns a grounded root-cause over the SDK path, **no MCP required**.
- [x] Optional Ollie hand-off (when the MCP is connected) produces the **same output shape** (documented; SDK path is primary).
- Open question from the ticket (does Ollie out-root-cause the repo-grounded agent for a single trace?) — the naive tests below confirm the repo-grounded SDK path is sufficient and strong; Ollie is positioned as a cross-check, not the default.

## Validation

**CI-clean:** passes the agentskills.io validator + pack build — `40 passed` (`tests/test_skills_spec_compliance.py`, `tests/test_skills_pack_build.py`); references resolve to `../opik/references/*`.

**Naive-agent tested 3/3** on **real traces** (agents given only the app + a trace id + this skill, no prior Opik knowledge):

| Case | Result | Root cause found | Read-only? |
|---|---|---|---|
| tool-failure | ✅ explained | `retrieve()` full-string lookup returns empty → ungrounded answer | no files touched |
| latency | ✅ explained | `fetch_context` `time.sleep(3.0)` = 3001ms of 3037ms | no files touched |
| bad id | ✅ not_found | clean 404, stopped at first blocker | no files touched |

All three used the SDK path (no MCP), returned the exact output contract with exactly one next step, and respected the read-only invariant.

## Notes

- Branched off `main` (has the pack CI + validator + metadata-fixed `opik`/`evaluate`). Independent of the still-open `/instrument` PR (#156).
- Reaches end users once the pack sync is enabled (gated on #156 per the cutover plan); the skill itself is independently mergeable.
- Evals harness (fixture-based, OPIK-7800-style) can follow as a fast-follow if we want regression parity with `/instrument`.


src/opik_mcp/skills/explain/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+Ask only at a genuine, non-inferable blocker (see **Blockers**).
+
+## Activation — the only in-scope work
+
+### 1. Resolve the target
+- A **trace id** (uuid-shaped): explain that one trace.
+- A **behavior/pattern** ("hallucinations since the prompt change", "slow responses"): search for the matching set (below), then explain the shared cause.
+- Confirm Opik is reachable: if `~/.opik.config` exists or `OPIK_API_KEY` is set, use it. Otherwise → **Blocker** ("run `opik configure`, then rerun").
+
+### 2. Fetch the trace and spans (SDK-first)
```
alexkuzmik: Can we invert the logic? First check if we have mcp available. If yes - use it. Otherwise SDK scripting is a fallback.

src/opik_mcp/skills/explain/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+- "This environment can't reach Opik — open the trace in the UI and paste its error/output, or run where Opik is configured."
+- "Which behavior should I explain? Give a trace id or describe what went wrong."
+
+## Output
+
+**User-facing:** a short human message — the root cause in one or two sentences, the evidence spans (names + why each matters), and the single next step. Not a raw trace dump, not JSON.
+
+**Underneath** (for composition / evals), one shape regardless of SDK-vs-Ollie path:
+- `status`: `explained` | `blocked` | `not_found`
+- `target`: `trace_id` (or the pattern + `trace_ids` sampled)
```
alexkuzmik: following similar comment in another review, let's keep the UX smooth when we can and give users clickable links, not just faceless IDs.

alexkuzmik (changes requested):

src/opik_mcp/skills/explain/SKILL.md
aswynz: Went the other way on this one, deliberately: rather than invert to MCP/Ollie-first, we dropped the optional Ollie reasoner entirely for now and kept the skill SDK-only (d4c25e9). Reasoning: the naive-agent tests showed the repo-grounded SDK path root-causes these traces well on its own (it has the code; Ollie does not), and OPIK-7649's open question — does Ollie actually out-root-cause the repo-grounded agent for a single trace? — is still unresolved. So there's no MCP-vs-SDK branch left to order. Happy to add an MCP-first path back once Ollie's added value is validated.

src/opik_mcp/skills/explain/SKILL.md
aswynz: Done (d4c25e9). The output contract now returns clickable Opik UI links (the trace redirect URL) for the trace and evidence spans, never bare ids, and the anti-patterns call out bare-id output explicitly.

src/opik_mcp/skills/explain/SKILL.md
aswynz: Update: reversed my earlier call and inverted it as you suggested (d624101). The skill now checks the hosted MCP first and uses it — including `ask_ollie` as the reasoner — when connected, and falls back to SDK scripting only when it is not. Restored the `reasoner` field (`ollie|agent`) and the Ollie references. Thanks for pushing on this.

PR #172, opened 2026-08-28
chore(mcp): stop advertising ask_ollie to the agent when it is disabled

aswynz:
## Why

`ask_ollie` is off in every shipped deployment (`OPIK_MCP_ASK_OLLIE_ENABLED` defaults to `false`; #170 removes it from the tool registry via `apply_tool_visibility`). But two **agent-facing** strings still advertised it, so the model is told to use a tool that isn't there:

- **Session instructions** (`instructions.py`) — the `InitializeResult.instructions` blob (the highest-leverage tool-routing text) had an unconditional `ask_ollie` bullet, plus mentions in the write and default-project guidance.
- **Entity description** (`read_list/registry.py`) — the `experiment` read/list description said "Pair with `ask_ollie` for analysis."

## What changed

- The `ask_ollie` guidance in the instructions blob is now **gated on `opik_mcp_ask_ollie_enabled`** — it renders only when the tool is actually advertised, matching what the server does. Kept as a gate (not a deletion) so it stays correct if the kill switch is ever flipped back on.
- Dropped the `ask_ollie` mention from the `experiment` entity description and from the write/default-project lines.
- Tests: the blob **omits** `ask_ollie` when disabled and **includes** it when enabled.

Implementation code (`ask_ollie.py`, analytics, `ollie_client.py`, config) is untouched — this is only about what the agent is *told*.

`make check`: 1378 passed, 2 skipped.

### DIFF
--- a/src/opik_mcp/instructions.py
+++ b/src/opik_mcp/instructions.py
@@ -6,8 +6,8 @@ Cursor, VS Code, Goose). Hosts that ignore the field lose nothing — each
 tool's description is self-contained, so this is purely additive.
 
 The content sets cross-cutting context (workspace, Opik URL, today's date)
-and primes the LLM on when to prefer ``read``/``list``/direct writes vs.
-``ask_ollie``. Per GitHub MCP's published data, dynamic per-session
+and primes the LLM on when to prefer ``read``/``list`` vs. direct writes.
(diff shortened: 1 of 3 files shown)


aswynz:
This is no longer relevant as PR https://github.com/comet-ml/opik-mcp/pull/181 addressed the issue

PR #173, opened 2026-08-31
[OPIK-8185] opik-instrument: verify span coverage + guard common ingestion traps

aswynz:
## What
Enhances the `opik-instrument` skill's **Step 6 — Verify ingestion** so that "verified" means a *complete* trace, not just that *a* trace arrived.

- **Coverage check** — compare the number of spans from `search_spans` against the call sites instrumented on the exercised path; every span must have a non-empty `name`/`type`, and LLM spans must carry input/output. Fewer or empty spans ⇒ not `verified`.
- **Common ingestion traps guard** — a short checklist for the failure modes that make a trace land empty / partial / unnamed: the batching race on fast spans, post-hoc feedback/score binding, and missing flush.

## Why
The skill already runs a safe path and checks that a trace arrives. But a trace can arrive *silently incomplete* — batching can drop or unname fast spans, and post-hoc scores can be lost if posted before the span's create reaches the backend. These were hit first-hand while instrumenting a multi-agent app from scratch: a tracer adapter that passed all unit tests and a static audit still lost trace data across three runs; only running the app and inspecting the platform caught it. A "code edits alone are not success" skill should catch that class of failure.

Docs-only change to the authored skill (`src/opik_mcp/skills/opik-instrument/SKILL.md`); the public `opik-skills` pack is built verbatim from it.

Closes OPIK-8185.

### DIFF
--- a/src/opik_mcp/skills/opik-instrument/SKILL.md
+++ b/src/opik_mcp/skills/opik-instrument/SKILL.md
@@ -85,6 +85,15 @@ To find the newest trace instead of using a known id, use the client's trace sea
 
 Traces are asynchronous — allow a few seconds after the run and make sure the flush ran.
 
+**Verify coverage, not just arrival.** A trace arriving is necessary but not sufficient — batching can silently drop or truncate spans, so a trace can land *incomplete* and still look fine. Before reporting `verified`:
+- **Count vs. expected.** Compare `len(client.search_spans(trace_id=tid))` against the call sites you instrumented on the path you ran (entrypoint + each traced tool/LLM). Fewer spans than expected means spans were dropped — do not report `verified`.
+- **Every span is well-formed.** Each span has a non-empty `name` and `type`; LLM spans carry input/output (and usage where the integration provides it). A span returned with an empty `name`/`type` is the batching-race symptom below, not a real span.
+
+**Common ingestion traps.** If the trace is empty, partial, or has unnamed spans, it is almost always one of these — not a bug in the instrumentation you added:
+- **Batching race on fast spans.** With batching on, a span created and ended within one flush window can be reordered by the backend and dropped (or stripped of its name). Ensure a single `flush()` at the very end and allow a few seconds before verifying.
+- **Post-hoc scores dropped.** A feedback score attached *after* a span closes (`log_traces_feedback_scores` / `log_spans_feedback_scores` by id) is lost if the span's create hasn't reached the backend yet — flush before posting the score, and reference the span by id.
+- **Missing flush.** The most common empty-trace cause: a script that exits without `opik.flush_tracker()` / `await client.flush()`.
+
 ### 7. Report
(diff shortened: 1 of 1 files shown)


src/opik_mcp/skills/opik-instrument/SKILL.md
```diff
@@ -85,6 +85,15 @@ To find the newest trace instead of using a known id, use the client's trace sea
 
 Traces are asynchronous — allow a few seconds after the run and make sure the flush ran.
 
+**Verify coverage, not just arrival.** A trace arriving is necessary but not sufficient — batching can silently drop or truncate spans, so a trace can land *incomplete* and still look fine. Before reporting `verified`:
```
alexkuzmik, line 88: @aswynz let's add a tip that if opik mcp is available - the agent can use it for verification.

alexkuzmik (changes requested):

src/opik_mcp/skills/opik-instrument/SKILL.md
aswynz: Added in e51726f: with the MCP connected, `read(entity_type="trace", id=tid)` inlines the span tree (up to 200 spans, `spansTruncated` flags the cutoff), so both coverage checks run over that one call. The SDK stays the default since it is already installed as part of instrumenting.

aswynz:
Refreshed rather than closed. I went looking for stale skill PRs and nearly closed this one on the strength of OPIK-8185 reading **Done** — but the content is not on `main`: no coverage-count or ingestion-trap text in `opik-instrument/SKILL.md`. The ticket is wrong, not the PR. (OPIK-8186, its eval harness, is MERGED, which is probably how 8185 got marked.)

Rebased onto `main` (clean) and fixed one thing the rebase surfaced: the coverage-count example called `search_spans(trace_id=tid)` without `project_name`, which is exactly the bug #191 fixed two lines above it — without it the SDK searches the configured default project and returns nothing. It now reuses the `spans` already fetched.

Ready for review. Skills tests pass and the pack builds 10 skills.

src/opik_mcp/skills/opik-instrument/SKILL.md
aswynz: Re-verified against `main` today, since `read` changed in #184/#187/#197 after this branch was written. The tip still holds exactly as worded:

- `read(entity_type="trace", id=tid)` — both parameter names are current (`server.read`).
- returns `{trace, spans, spansTruncated}` with the tree inlined, `SPANS_INLINE_LIMIT = 200` (`read_list/entities/trace.py:30`), so the 200 figure and the `spansTruncated` fallback are right.
- #197's new `fields` defaults to `None`, so the default read still inlines the spans — the one call covers both coverage checks with no script.

The SDK stays the default in the skill only because it is already installed as part of instrumenting; the MCP is the shortcut when it is connected. Resolving — happy to reopen if you'd rather the MCP lead instead of follow.

awkoy (approved):

PR #176, opened 2026-09-01
Keep eval fixtures out of the published wheel

alexkuzmik:
Found while checking why `read_skill` was missing from PyPI. **This would have shipped for the first time in the next release.**

## The defect

`opik-diagnose` and `opik-explain` carry `evals/` fixture directories. Three places decide what ships, and only two of them exclude those:

| surface | excludes `evals/`? |
|---|---|
| MCP resources + `read_skill` | yes — `skills_catalog.EXCLUDED_DIRS` |
| published `opik-skills` pack | yes — `build_skills_pack.py` imports the same rule |
| **the wheel** | **no** |

Building the `0.2.27` tag gives 39 skill files, of which the server will ever serve 21. The other 18 are 41 KB of fixtures, including runnable `agent.py`, `seed.py`, `grader.py`, `run_evals.py`, landing in every user's `site-packages`.

`0.2.24` on PyPI is clean only because it predates both skills — so nobody has seen this yet, and nobody will if it goes in before the next release.

## The fix

One `exclude` in `[tool.hatch.build]`. Wheel goes 307 KB → 286 KB, 39 skill files → 21.

## Why no test caught it

Nothing in the suite looked at a built artifact. `test_skills_packaged.py` resolves skills through `importlib.resources`, which in a dev checkout reads the **editable** install pointing at `src/` — it passes whatever the backend would or would not put in a wheel.

`tests/e2e/test_wheel_contents.py` builds a real wheel with the real backend and reads it back, in both directions: every served document is present, and nothing that will not be served is. Plus the end of the chain — install the wheel into a throwaway venv and have *that* interpreter enumerate the skills, which is the only thing a pip-installing user actually has.

Verified by removing the exclusion: two tests fail, exactly the state that would have shipped.

`e2e`-marked (shells out to `uv build`, ~3s), so it runs in the existing `e2e` CI job.

## Release note

`0.2.27` is the tag carrying `read_skill` (merge commit `2b8ac10`). If this merges first, release the tag that lands after it instead, and PyPI gets both the new tool and a clean wheel.


aswynz:
Anything needed before this can be merged @alexkuzmik  ?

PR #186, opened 2026-09-09
[OPIK-8310] Diagnostics disabled: explain the empty issue list, and let the agent turn it on

awkoy:
## Details

#184 gave the MCP the Diagnostics issue list but left the empty case alone. "No agent_insights_issues found." meant four different things an agent could not tell apart: Diagnostics unavailable on the deployment, never enabled for the project, turned off, or on and clean. Only the last one means nothing is broken, so the agent reported false all-clears. In two of the four it could have fixed the situation itself, because the backend lets the credential that reads issues also create the project's job and trigger a scan.

An empty issue list now names its state and, where something can be done, the call to make and the project's Diagnostics page. If the job recorded a failure, it quotes the reason. A failed lookup falls back to the plain message, so the hint can never turn an answered list into an error.

A non-empty list dates itself. Its issues are whatever the last scan grouped, and no row said when that was, so a report stopping at yesterday read as the state of the world now. Ask for a week on a nightly-scanned project and the last day was missing from the answer with nothing to say so. The reply now ends with what it covers, and when the requested window runs past that, with the size of the uncovered tail and the way to close it: a trigger when a rescan reaches back far enough, otherwise raw traces with the `since` it names. A trigger only covers 24 hours, so on a wider gap it would leave the middle missing while reading as the fix.

Two write operations for the job, both taking `project_id` or `project_name`, both single-item, both scoped to `project_data_view` (what opik-backend requires on every job endpoint, and what reading issues already needs):

- `agent_insights_job.enable` turns Diagnostics on. On the backend's 409 it flips the existing job back to enabled, so calling it twice is safe.
- `agent_insights_job.trigger` scans the last 24 hours now and returns the page link, instead of waiting for the 00:05 UTC run. A 404 means no job yet, so it says to enable first.

Three more for an issue's lifecycle, `agent_insights_issue.resolve`, `.close` and `.reopen`, over one backend route that has been there all along. Reading issues without being able to retire one leaves a growing pile: a failure fixed last week still reads as open and gets ranked again. Each answers with a link to the view the issue moved to, since a resolved issue is no longer on the default page. The descriptions and the skill both say these are for when the user asks. Whether a failure is dealt with is a judgment call, and an agent tidying the list while triaging would be hiding work.

A deployment gate guards the two job operations. Without Ollie the endpoints still answer: a job can be created and a trigger is queued to a stream nothing reads, so the scan never runs and never fails. The gate reads the backend service toggle the UI reads for the same decision, and a missing field or a failed read means available, so older backends behave as today. The issue operations skip the gate, because an issue that exists is proof Diagnostics runs here.

`opik-diagnose` reads the state instead of assuming an empty list means "nothing grouped yet". It triggers a stale scan on its own, asks once before enabling (a standing daily job is the user's call), never waits for a scan, and acts on the coverage line rather than presenting a dated report as the whole window.

Driving the skill by hand turned up two things its conformance tests cannot see:

- It requires a clickable link per shortlist item and forbids bare ids, but nothing on the MCP path handed one over. `list('trace')` has no url column, `read('trace')` no url field, and the only link source was an issue read's `trace_url_template`. An agent had to invent the shape, and a guess gives a plausible 404. The session instructions now name the template once, using opik-backend's redirect (the link the Python SDK prints) instead of a direct project URL: the redirect resolves the project and the workspace from the trace id, so the agent fills one slot, needs no `project_id`, and gets a correct link even where the workspace is unknown.
- `source = "SDK"` returns a 500 with nothing but a log id, because the backend deserializes the filter value into its enum and throws, and the throw arrives as a server error instead of a 400. Closed enum filter values are now checked before the call, every element of an `in` list included, and `schema("list.<entity>")` lists them. Filed as OPIK-8320. The value sets were checked against a live backend, which corrected two of them: `unknown` is legitimate, and span `type`, thread `status` and `visibility_mode` are not validated server-side at all, answering with a silent empty page.

Two smaller things. `read`/`list` accept `issue` for `agent_insights_issue`, unadvertised, as a net under the guess an agent makes anyway. And the toggles cache is gone, which reverses a call made while grilling: the gate is consulted only for Diagnostics work, so it was trading a live answer for savings nobody would measure, and an operator switching Ollie on mid-session now sees it at once.

Out of scope, as agreed: polling for a scan, a `read('agent_insights_job')` entity, `disable`.

## Change checklist
- [x] User facing
- [x] Documentation updated (if needed): README, instructions blob, tool descriptions, skill
- [ ] Breaking changes documented (if any): none. Five new write operations; `read` and `list` schemas untouched.

## Issues
- OPIK-8310 (follows OPIK-8285 / #184). Filed from this work: OPIK-8320.

## Testing

- `make check`: 1605 tests, mypy --strict, ruff. `write.json` and `schema.json` regenerated for the five operations; `read.json` and `list.json` unchanged.
- Ran the whole Diagnostics lifecycle through real MCP tool calls against prod, on a disposable project: the empty list said "not enabled"; `enable` returned 201; a second `enable` went 409 then PATCH 200; `trigger` returned 202; once the scan finished the list read `No open issues. Last scan: 2026-09-09T12:11Z.`; and a REST `PATCH status=disabled` moved it to "turned off". All five states seen on the wire.
- Followed the `opik-diagnose` skill by hand as an agent, which is what surfaced the missing trace link.
- Opened a rendered trace link in a browser: the redirect resolved to the trace's project and workspace and opened the trace panel. A template rendered for one workspace opened a trace in another, which confirms the workspace comes from the trace and not from the session.
- Ran the `opik-diagnose` eval harness. The functional case passes with recall, precision, signal coverage, read-only and schema compliance all 1.0, driven through the MCP against a freshly seeded project. `selection_accuracy` is 1.0, but read that one with care: the judge panel was me, and I knew the expected answers. It also had to be scored twice, because the grader read the judge's verdicts at the wrong nesting level and reported 0.364 for a correct sheet (fixed here).
- Reviewed on two axes, and every finding applied. The one correctness bug it found: the coverage note measured the gap from the window's end but described it from now, so with `until` in the past it named the wrong stretch and offered a trigger that could not reach it. The empty path had the same bug. Both now judge staleness against the window's end and bound the suggested call at both ends. The old tests only used an unbounded window, which is how it got through. The rest were structural and are in: two `if entity_type ==` branches became one registry hook (`page_note_fn`, the counterpart of `link_fn` for lists), `run_write` handed its Diagnostics work to three stage helpers and went from 165 lines to 96, and three duplicated shapes collapsed into `_read_job`, `_RequiredProjectMixin` and `_refuse`.
- Two defects the live runs found and this branch fixes: right after `trigger`, the hint told the agent to trigger a scan it had just started (the job record has no in-flight field, so the wording now covers both cases), and the hint rendered its payload as Python kwargs instead of JSON.

## Review round 2

@alexkuzmik asked for the abstractions to be cleaned up, and named `dispatch.py` holding operation-specific branching, with a note that Diagnostics was not the only offender. He was right.

An operation's registry entry now carries what that operation does differently, as hooks: `validate_fn`, `build_fn`, `prepare_fn`, `retry_fn`, `decorate_fn`, `dry_run_note_fn`. Same pattern `read_list`'s registry already uses for `link_fn`. The behaviour moved to `writes/operations/`, one module per family next to the models, and the hook signatures plus the shared wire helpers to `writes/wire.py`.

`dispatch.py` went from 868 lines to 324, the thirteen-branch `_build_request` chain is gone, and it names no operation at all. That last part is a test, not a promise: `test_dispatch_stays_generic` searches the module for every registered operation name, pins that every hook comes from an operations module, and pins that no operation silently falls back to the default body. The 213 write tests are untouched.

The two Diagnostics modules in `read_list` became a `read_list/diagnostics/` package, also as asked.

On the SDK's OQL: looked at it and answered in the thread. Our grammar tests were ported from its parser tests, so they agree on syntax, but `opik-mcp` has no dependency on the SDK (its install pulls litellm and boto3-stubs) and the SDK has no accepted-values table, which is the part that turns `source = "SDK"` from a 500 into a 400. A small shared package for the field and operator tables is the honest fix if we want one source of truth; happy to file it.

## Documentation

README (the Diagnostics paragraphs, filters, the write-operations table, trace links), the instructions blob, the five operation descriptions, and the `opik-diagnose` skill.


src/opik_mcp/writes/dispatch.py
```diff
@@ -93,11 +97,25 @@ async def run_write(
     # thread's model UUID; the caller passes the thread_id string). dry_run stays
     # pure — no client, no backend calls — so it never depends on config/network.
     http_client: OpikClient | None = None
+    resolved_settings = settings or get_settings()
+    diagnostics_project_id: str | None = None
```
alexkuzmik: This generic dispatch.py contains too much diagnostics-specific logic and branching. Ideally, it should remain as generic as possible for write operations (just be a dispatch). And all the specifics for write operations should be encapsulated into proper entities to not overcomplicate the dispatching logic.

This is relevant for other logic in this module too, I guess diagnostics is not the only one.

src/opik_mcp/read_list/diagnostics/availability.py
```diff
@@ -0,0 +1,71 @@
+"""Does this deployment have Diagnostics at all?
```
alexkuzmik, line 1: if it's related to diagnostics, we can move both of them to the following sub-namespace.

src/opik_mcp/read_list/oql.py
```diff
@@ -536,6 +574,19 @@ def _validate_value(field: str, ftype: str, key: str | None, raw: _RawClause) ->
         return OQLIssue(
             "bad_value", f"Empty value for '{field}': the backend rejects blank filter values."
         )
+    allowed = ENUM_VALUES.get(entity_type, {}).get(field)
+    if allowed is not None:
```
alexkuzmik, line 578: Just FYI we have an OQL in Python SDK, might be worth taking a look.

alexkuzmik (changes requested):
The logic looks fine, I'd only ask to make a few more iterations to cleanup the logic and abstractions a little. Code cohesion and coupling are the things that will matter for effecient AI development as the code base grows

src/opik_mcp/writes/dispatch.py
awkoy: Agreed, and you were right that Diagnostics was not the only one. Done in 6576fe3.

An operation's registry entry now carries what that operation does differently, as hooks, the same pattern `read_list`'s registry already uses for `link_fn`:

- `validate_fn` — a rule the model cannot express (a score batch must be homogeneous by target; `target='thread'` has no singleton route)
- `build_fn` — payload to request. Defaults to the endpoint plus a dump of one item, so only the operations with a batch envelope, a path id or a renamed field need their own
- `prepare_fn` — resolve an identifier the wire needs but the caller does not carry, or refuse. A thread comment's model UUID; a Diagnostics project
- `retry_fn` — reinterpret a backend answer that means something other than what it says
- `decorate_fn` / `dry_run_note_fn` — what to add to the result, and what a preview cannot show

The behaviour moved to `writes/operations/`, one module per family next to the models: `observability` (traces, spans, scores, comments), `evaluation` (prompts, test suites, experiments), `threads`, `diagnostics`. Hook signatures and the shared wire helpers are in `writes/wire.py`.

`dispatch.py` went from 868 lines to 324, and the thirteen-branch `_build_request` chain is gone with it. It now names no operation at all, and that is a test rather than a promise: `tests/test_writes/test_dispatch_stays_generic.py` searches the module for every registered operation name, because this is the sort of boundary that erodes one convenient branch at a time. It also pins that every hook comes from an operations module, and that no operation silently falls back to the default body.

The 213 write tests are untouched and pass, which is the only reason a move this size was worth making.

src/opik_mcp/read_list/diagnostics/availability.py
awkoy: Done in 6576fe3. Both files are now `read_list/diagnostics/`: `availability.py` (can this deployment scan at all?) and `state.py` (what is this project's Diagnostics doing, and what does a page of its issues not say for itself). The package docstring says why they sit together: they are read together, an empty issue list asks both, and neither is any use to an entity other than `agent_insights_issue`.

The write side got the same treatment in the other thread, so the operations that turn Diagnostics on and move an issue are `writes/operations/diagnostics.py`.

src/opik_mcp/read_list/oql.py
awkoy: Thanks, I looked at it: `sdks/python/src/opik/api_objects/opik_query_language.py`, 826 lines, an `OQLConfig` per entity with the operator maps per field type. Our grammar tests here were ported from its parser tests for exactly that reason, so the two agree on syntax by construction.

Two things stop me from reaching for it directly:

`opik-mcp` does not depend on the `opik` SDK at all. Its seven dependencies are mcp, httpx, pydantic, pydantic-settings, sentry-sdk and uvicorn; it talks to the backend over REST itself. The SDK's install pulls litellm, boto3-stubs and rapidfuzz among others, which is a lot of surface to take on for a parser, in a process that starts on every MCP session.

And it does not do the part this change is about. The SDK's configs stop at field to operator; there is no table of accepted enum *values*, which is what turns `source = "SDK"` from a 500 into a 400 here. Its entity coverage also differs: it has dataset_item and prompt_version, we have experiment.

Where they genuinely overlap is the field and operator tables, and there we can drift. If you want one source of truth, the honest shape is a small shared package that both import, rather than one side taking the other's dependency. Happy to file that as a follow-up if you agree it is worth it.

alexkuzmik (approved):

PR #187, opened 2026-09-10
[OPIK-8284] Project overview in read('project'), and one metric over time in list('project_metric')

awkoy:
## Details

"How is my project doing" had no answer here. An agent asked it by listing traces and counting them, which costs a page of rows to produce four numbers, and the numbers still disagreed with the ones on the user's screen, because the Logs cards count SDK traffic only and compare against the period before.

`read("project", …)` now answers it in one call. Alongside the record it returns the four figures the Logs page shows as cards, trace count, error rate, average duration and total cost, for the last 7 days against the 7 before, filtered to `source = "sdk"` as the page is. It comes from the same `kpi-cards` endpoint the UI calls, so the numbers match the screen. `since` / `until` move the window, and the reply states the window it used and the period it compared against, since "compared to last month" when it was last week is the mistake this is cheapest to prevent. A whole project read measures 254 tokens against production.

An empty period returns a zero count, a zero cost and a zero error rate, and the last of those is a lie worth catching: 0% errors on a week with no traffic reads as a healthy week, and someone may act on it. A rate or an average over a period with no samples comes back as `null`, with a note saying why. Counts and sums stay zero, because zero is what they are.

The read also carries the project's vocabulary: its feedback score names, its token usage keys, and the automation rules producing most of those names. Those names are what the rest of the tool is driven by. Filters and groupings are written against names an agent cannot guess, and a guessed name returns an empty page that reads like good news. Score names and rules are capped by count, state the true total, and name the `list` call that returns the rest; usage keys are returned in full, because no call enumerates them on their own and a cut there had no way back. Last, `contains` names the freshest experiment, test suite, prompt version and optimization run, which is the cheap version of "what has been happening here".

The question that follows an overview is when something changed, so `list("project_metric", …)` charts one metric over time: twenty of them across traces, spans and threads, bucketed hourly, daily, weekly or into one total, filtered with the same OQL as every other list. It rides on `list`; a sixth tool was the alternative. Every tool in `tools/list` sits in the context of every request a host makes, including the ones that never touch Opik, so a tool is a permanent tax on people who may never call it. That is ADR 0004's "one schema covers N entities", and the cost of the alternative was measured. The series is a pipe table for the same reason: the same 30-day answer with ten groups is 3,649 tokens as nested JSON and 871 as a table.

When the caller names no interval, it follows the window the way the Metrics tab does: hourly up to 3 days, daily up to 30, weekly beyond. So `since="1h"` charts by the hour instead of returning one daily bucket, the 7-day default stays daily, and a year is 52 rows rather than 365. The header echoes a derived interval as `daily (from the window)`, the same way it echoes a defaulted series, because the numbers depend on a choice the caller never typed. An explicit interval is sent as given: an hourly month is 721 rows, and whether that is worth the context is the caller's decision. An earlier version refused any request over 200 buckets before calling the backend. The number was ours, no layer of Opik has one, and a refusal turns "this answer is large", which the caller can see and narrow, into "the MCP would not answer".

`breakdown` splits each bucket by tags, name, error info, error type, model, provider, span type, guardrail name or a metadata key. Which metrics accept which is transcribed from the backend's own `BreakdownField` instead of discovered by trial, because its refusals cannot be acted on: ask to group `span_cost` by model and it answers "This field supports Span metrics only", about a span metric. The cause is that its compatibility sets omit seven of the twenty metrics, so those seven support no grouping at all (OPIK-8336). Until that is fixed on the backend, this refuses locally with the real reason and, where one exists, the metric that does answer the same question. Where none exists it says so, because a refusal that only says no costs a second refusal: told that `trace_cost` cannot be grouped by model, the obvious next move is `span_cost`, which is in the same omitted set.

Three metric families come back as several series at once (a duration as p50/p90/p99, a feedback score per name, token usage per key), and the backend cannot fan those out and group them at the same time. It requires a sub-metric instead, which left grouping unreachable for eight of the thirteen groupable metrics: the answer was a 422 saying `sub_metric is required` and there was no argument to comply with. Hence `series=`, one argument for all three families, since at most one applies at a time. Duration defaults to p50 and token usage to `total_tokens`, both echoed on the first line so a default is never silent. A feedback score gets no default, because the names belong to the project and picking one would answer a different question without saying so. A name the project does not record is refused with the names it does record, and only when the answer came back empty, so a chart with numbers in it pays nothing for the check. Ungrouped, a score or usage metric returns every series it has, one column each; an eleven-column cap used to keep the widest and drop the rest, and nothing could chart the twelfth.

Four things the backend does that this has to correct, all confirmed in its source:

- The grouped queries have no `WITH FILL`, so each group carries only the buckets it appeared in and groups come back different lengths. Read by position, a model that ran on one day of seven lands in the first row under somebody else's date. Rows are keyed by timestamp.
- `__others__`, the bucket for groups past the backend's tenth, is a concatenation, not an aggregate, so one series can hold several points at one timestamp. They are summed where the metric is additive and left blank where a sum would not be that metric, and the table says which.
- The metrics endpoint filters thread metrics with a field set that has neither `source` nor `environment`, and drops what it cannot apply without a word. Thread metrics no longer add the SDK default (the header would have claimed a filter that never reached the database), and naming either field on one is refused.
- A feedback score average of exactly 0 comes back as `null`, so an all-empty score table says no bucket carries a score, where it used to claim every bucket was zero.

### Nothing is cut here any more

`read` used to shrink what it returned: every string over 200 characters truncated and stamped with a jq path, and a composite too wide for that reduced to a skeleton of ids. That came from ollie-assist, where it is safe, because ollie caches the whole entity for the session and ships a jq tool, so the pointer it leaves is an instruction the agent can carry out. We took the truncation and left the cache and the tool behind. The pointer named a place that does not exist and the data was gone; the only way back was a second read with a larger budget, which nothing said. A large answer costs context the caller can see; a silently short one costs a wrong conclusion they cannot, and the report that follows says "the MCP is wrong", not "the MCP truncated my trace".

So the compression module, its three tiers, `max_tokens` and the per-entity compressors are gone. A read returns the record whole, with its size stated on the first line.

The children a composite read inlines are the one exception, and they use the backend's truncation. `?truncate=true` on the trace, span and thread list endpoints reads ClickHouse's materialised `truncated_input`/`truncated_output` columns instead of the raw ones, so a field over 10 KB never comes off disk, and base64 images become the literal `"[image]"`. What makes it safe is the half ours lacked: `GET /traces/{id}` and `GET /spans/{id}` take no such parameter, so every child carries the id that fetches it back whole, and the threshold is the product's own (`ResponseFormattingConfig.truncationSize`, the figure the UI's tables use). The backend does not report its own cut, its `input_truncated` flag never reaches the API, so the read counts what arrived cut and says so in `spanBodies` / `messageBodies`, naming the call that returns one child whole. A thread slims its own record too: `first_message` is `argMin(t.input, t.start_time)`, the same bytes `messages[0].input` carries, and shipping one payload twice at two lengths is worse than either choice made consistently.

The same rule was then applied to every other place the server returns less than exists. An inlined collection longer than 200 spans, 200 turns or 100 versions used to set `spansTruncated: true` and nothing else; beside it now sits `moreSpans` / `moreMessages` / `moreVersions` with the count and the exact `list(...)` call that continues from where the inlined part stopped. That call has to return the same set, so `list` no longer adds its `source = "sdk"` default when the filter names a `trace_id` or `thread_id`: the default exists for triage, a filter on a parent's id is a drill-in, and for a trace from an evaluation or the playground the default made the continuation return a different set than the part it continued. A name that matched more than ten candidates lists ten and says how many more.

### Layout

Review asked for one namespace per entity, and the read/list package now has it: each entity is a module or package under `entities/`, ending in the `EntityHandler` the registry registers; the root holds only the two dispatchers, the handler contract, and what two or more entities share (the filter language, the window vocabulary, paging, project scope, the slim notice). A test pins the layout, since a property nobody checks erodes on the next ticket.

`tools/list` measures 20,314 bytes. A test caps it at 24,000, with a history of who spent what. The cap used to sit ~400 bytes above the measurement and was hit three times inside this ticket, each time trading a sentence the agent needed for a handful of bytes; it now guards against the multi-kilobyte accident (the `write` entry grew 1.9 KB in a PR nobody measured) and leaves wording alone. Reference material stays behind `schema("list.project_metric")`, paid only by a caller who is about to chart something.

Out of scope, as agreed: workspace-level metrics, alerts, and any write path.

## Change checklist
- [x] User facing
- [x] Documentation updated (if needed): README (the project read, the metric series, the two name lists, what is and is not cut), instructions blob, tool descriptions, `schema("list.project_metric")`, the `opik` and `opik-diagnose` skills
- [x] Breaking changes documented (if any): `read` loses its `max_tokens` argument, and no longer truncates. `read("project", …)` gains keys and keeps the ones it had; `list` gains three entity types and three optional arguments.

## Issues

- OPIK-8284. Ten tickets under `~/.claude/work/opik-mcp/project-overview/issues/`, worked in dependency order.
- Filed from this work: OPIK-8336 (`BreakdownField` omits seven metrics and the resulting 422 contradicts itself), OPIK-8337 (`/projects/{id}/activities` takes 5 to 11 seconds on a project with an optimization run), OPIK-8338 (some `Demo evaluation` traces carry UUIDv7 ids stamped year 2201, so they pass every `since` window, and `KpiCardDAO` binds on the same id bounds, so the summary inherits it).

## Testing

- `make check`: 1,760 tests, mypy --strict, ruff. `make e2e`: 30 tests over stdio against an HTTP stub of opik-backend, covering the whole surface this ticket adds, including what reaches the wire (`truncate=true` on the spans request and its absence on `GET /traces/{id}`; the metric request's interval, filters and sub-metric).
- Built the wheel, installed it, registered it in this Claude Code session and drove the whole surface through real tool calls against www.comet.com, as a user would. Three defects came out of that pass and none of them from the suite: rate and average series printing 0 for buckets with no traffic; two refusals needed to learn one fact; and grouping a token metric being unreachable at all. A fourth came out of reading the output afterwards: a `series` name the project does not record charted nothing and looked like a quiet window, which was the one place a wrong argument produced a plausible answer. Repeated after the truncation change: a trace read against production returns its span with the body intact and the notice beside it.
- Reviewed in parallel on two axes (repo conventions and smells; fidelity to the tickets and to the design decisions taken during the work), four times over the life of the branch, the last over the whole diff. Findings that changed the design rather than a line: the misattributed grouped series; the thread filter that never applied; the thread read shipping one payload at two lengths; the slim notice, which began as a fixed sentence and became a count derived from the payload, because a notice that reads the same whether or not anything was cut cannot be checked against the answer beside it; and the continuation call above, which named a set the default filter then changed. A test now runs that call as written and checks it returns exactly the spans the read left out.
- The test review was mutation-based. It found the "one clock reading" regression test asserting two values the code had handed itself: it passed with the fix deleted. It now freezes the read's clock so a second reading would land years away, and removing the fix turns it red. The metric request also goes over a mocked transport through the real client, so an argument the client has no parameter for fails here instead of in production. The read fakes were changed to honour `size` and report the true total, since a fake that returns everything can never produce a short page, and the pointer tests would have been untestable against it.
- Verified live after each fix, including the ones that are only visible in production data: a sub-cent cost that used to print as `0` now reads `1.35e-05` and matches the same figure in the project summary.


src/opik_mcp/read_list/project_contents.py
```diff
@@ -0,0 +1,99 @@
+"""What a project contains, distilled from its activity feed.
```
alexkuzmik: All project_ modules can be moved to a dedicated namespace.

src/opik_mcp/read_list/registry.py
```diff
@@ -437,6 +483,40 @@ async def _list_projects(client: OpikListClient, **kw: Any) -> dict[str, Any]:
     return await client.list_projects(**kw)
 
 
+async def _list_score_names(client: OpikListClient, **kw: Any) -> dict[str, Any]:
```
alexkuzmik: /goal Let's refactor `read_list` namespace in a way that every entity and all the logic corresponding to its processing is stored in its own sub-namespace.
The root read_list modules should ideally contain the dispatching, registering, or just universal logic relevant for all entities, but not huge files containing a ton of helpers for anything. The read_list, so as any other tool with a registry of supported entities, should be modular. Ideally, whenever we add a new entity, we should add a namespace with its logic and then just import + register it.

alexkuzmik (changes requested):

src/opik_mcp/read_list/registry.py
awkoy: Done — `read_list` is modular now, same shape as `writes/` after #186.

`registry.py` went from 995 lines to 93. It imports thirteen handlers, keys them by entity type, and keeps only the two behaviours that read the table rather than a row: the alias an `entity_type` resolves through, and the compressor an entity did not override. About 900 of the lines it lost were thirteen entities' fetchers, list functions, name searches and compressors.

Each entity now declares its handler beside its own code under `entities/`: a module where one file is enough (`span`, `experiment`, `prompt` with its versions, `test_suite` with its items, `trace`, `thread`, `score_name`, `online_rule`), a package where it is not.

- `project/` — the read that fans out, plus `summary`, `vocabulary`, `contents`, one file per part of the answer.
- `project_metric/` — `catalog` (what can be asked, and every refusal that needs no call), `runner` (the order of operations), `table` (how an answer reads), `reference` (what `schema()` says).
- `agent_insights_issue/` — the Diagnostics package moved under the entity it belongs to, with that entity's read and list joining `state` and `availability`.

Three things came out to the root because more than one entity needs them: `handler.py` for the contract, `paging.py` for reading the page envelope, `unsupported.py` for the two sentinels that belong to no entity. The contract's move also removes a circularity you may have noticed — `diagnostics/state.py` was importing `PageContext` from the registry under `TYPE_CHECKING`.

The dispatchers no longer name an entity. The two remaining `entity_type ==` branches became hooks on the handler, the way the write registry carries `validate_fn` and friends: `run_fn` for an entity that answers `list` whole (a time series is not a collection), `reference_fn` for one whose schema is a catalog instead of a field table.

`test_modular.py` pins the property so the next ticket cannot quietly undo it: the root's module list, a namespace per registered entity, a registry that defines only its two table-level functions and nothing async, no entity name in either dispatcher, and the readable/listable surfaces unchanged by the move.

Behaviour is identical — the handler literals were lifted from the registry's own source rather than retyped, so no entry lost a column, a flag or a description. 1756 tests, ruff and mypy clean, and re-verified live through the packaged wheel.

c7af94a

awkoy:
One more push on top of the namespace split — I reviewed it myself before handing it back, on the same two axes, and four things came out. Worth knowing before you look, since two of them changed the shape again.

- **A message I broke in the refactor.** Generalising the metric runner turned `list('project_metric', …)` into `list(project_metric, …)` in the timeout error. The quotes are the difference between a call form an agent copies and one it cannot. Restored, and the collection path now quotes its entity too, which it never did.
- **The window reason had moved the wrong way.** The same commit replaced two `entity_type ==` branches with hooks, then put a third (why an experiment takes no window) into a dict in `oql.py` keyed by entity name. It is `EntityHandler.no_window_reason` now, declared by the entity it is about.
- **`delegated_elsewhere` is gone.** `project_metric` declared a `list_fn` sentinel purely so `list_fn is not None` would make it listable. `EntityHandler.lists` asks whether either route answers, so the handler no longer declares a function nothing calls.
- **One entity was importing another.** The metric runner reached into `entities/project/vocabulary` for the names it validates `series` against. Both were root modules before the split, so the split caused it. Which score names and usage keys a project recorded is a fact two entities need and neither owns; it is `read_list/project_names.py` now.

`test_modular.py` changed with them. It had hardcoded both surface lists, so adding an entity meant editing the test — the opposite of the property it is guarding — and its headline check compared filenames, which is why the cross-entity import above walked straight past it. It now pins: no entity imports another, every entity module declaring a handler is registered, an entity with a `run_fn` declares no `list_fn`, and the contract imports no entity.

Two things I left alone, with the reasoning in the commit rather than argued here. `run_fn` is thinner than a hook looks (one runner, and the dispatcher still passes the metric's own kwargs) — its value is that the choice became registry data a test can pin, not reuse that does not exist yet. And the per-entity tables in `oql.py`, `sorting.py` and `uri.py` stay at the root: they are one grammar, and splitting it across thirteen files costs more than it buys. Happy to move them if you read that line differently.

1758 tests, `make check` clean, re-verified live through the packaged wheel. b393ba8

alexkuzmik (approved):

PR #188, opened 2026-09-14
[OPIK-8369] list('test_suite_item'): columns from the items' data keys, cuts declared

awkoy:
## Details

`list('test_suite_item', test_suite_id=…)` returned every row empty. The projection was hardcoded to `input` and `expected_output`, two fields a dataset item does not have: its payload is `data`, a map keyed however the user built the dataset (`question`/`answer`/`expected_behavior` on the suite this was reported on, `prompt`/`completion` on the next). A fixed two-column projection could never hit them, so twenty rows of `id | name | input | expected_output` came back blank while the same content was plainly visible on the traces that ran against those items.

The columns are now read off the page. The handler declares a `list_projection_fn` (new on `EntityHandler`, beside `list_extra_fields`) that unions the `data` keys across the returned rows and ranks them: the SDK-documented keys first (`input`, `expected_output`, `output`, `context`, `reference`) when a dataset has them, then by how many rows on the page fill the key, then by name, so the same page always gets the same columns. Nothing is hardcoded to any dataset's vocabulary; the ranking only decides order.

Every cut the table makes is stated under the rows, in the voice of the existing headers, per the invariant in `size.py`:

- A union wider than eight keys is cut, and the note names what was left out: `Columns after id are the items' data keys, showing 8 of 12 by fill rate; omitted: source_url, ticket_id, verified, version.`
- Values share a per-page character budget, split across rows × columns into a per-cell cap, floor 60, ceiling 4000. Twenty rows of three columns get 133 chars each; `size=1` gets a value whole. The line says so: `20 values cut at 133 chars; fewer rows per page (size=…) raise the cap.` That is the drill-down for a long value: an item has no read of its own and the server keeps no copy of the page, so the way to see more is to ask for less.
- Nested values render as compact JSON, not Python's repr and not `[object]`. This also changes how a `tags` list renders on a trace row: `["prod","beta"]` instead of `['prod', 'beta']`.
- The cut line is not specific to items. Any list that truncated a value at the 60-char default now says `1 value cut at 60 chars.` under the table. Before, the cut was silent.

The always-empty `name` column is gone from item rows (`list_has_name=False`).

The handler's description claimed `read('test_suite')` inlines up to 200 items. It never did, and items belong in the list: a suite read is ~200 tokens and the items are the bulk. The sentence now says what the code does.

**Why not the backend's column endpoints.** The items page already carries `columns`, the whole dataset's key union with types, at no extra call; but it cannot say which keys *this page's* rows fill, which is what the ranking needs. `/items/experiments/items/output/columns` without `experiment_ids` enumerates experiment *output* keys, not item data keys. Both are noted in `project_items`' docstring; the union is taken from the rows.

Out of scope, deliberately: the experiment comparison grid, an `experiment_item` entity, `dataset_item_id` on the trace projection, any caching or snapshot.

## Change checklist
- [x] User facing
- [ ] Documentation updated (if needed) — `docs/` is gitignored in this repo; the stale `design.md` row was fixed locally only
- [ ] Breaking changes documented (if any) — none; the `list` input schema is unchanged, conformance snapshots untouched

## Issues
- OPIK-8369

## Testing

`make check` passes (lint, mypy strict, 1784 tests).

New: `tests/test_read_list/test_test_suite_items.py`. Varied key sets, the SDK keys leading, missing keys ranking by fill, `null` counted as unfilled but keeping the column, the eight-column cap with the omitted keys named, the cell cap scaling with rows × columns, long values cut and declared, a smaller page showing the value whole, nested values as compact JSON, the empty page, rows without a `data` map, and the handler description matching the code. `test_list_tool.py` pins the generic cut line and its absence when nothing was cut.

Live, against production through the built wheel installed as a local MCP server:

- `andreicautisanu` / suite `019f8d97-c83c-7597-b40a-bd2e0e1ad558` (the reported one): all 20 rows show question, answer and expected_behavior. The hosted `opik-prod` on the same call still returns blanks.
- `awkoy-v2`, four fixture suites created for this (`mcp-8357-ragged`, `mcp-8357-nested-long`, `mcp-8357-wide`, `mcp-8357-empty`): six rows with different key sets and nulls rank by fill with empty cells where a key is absent; nested `input`/`expected_output` render as JSON, a 1400-char clause is cut at 533 on a 3-row page and whole at `size=1`; twelve keys cut to eight with the four omitted named; the empty suite answers `No test_suite_items found.`
- Two existing datasets with other vocabularies (`_generated`/`_generation_model`/`how`; `input`/`expected_output`/`metadata`) render their own keys.

## Documentation

None shipped; `docs/` is not tracked here. The behaviour is described in the module docstrings of `list_tool.py`, `handler.py` and `entities/test_suite.py`.


awkoy:
Closed by the branch rename to OPIK-8369 (the real ticket). Superseded by #189, same commit.

PR #192, opened 2026-09-17
[OPIK-8396] Show the full experiment record in list('experiment')

awkoy:
## Summary

Five tickets from the experiment-flows epic, plus one item from a sixth. They share two files, the list tool and the comparison layout, so they land together and the files are merged once.

| Ticket | What changes |
|---|---|
| OPIK-8396 | `list('experiment')` shows the full record and takes `type`, `optimization_id` and `experiment_ids` filters |
| OPIK-8394 | The comparison table signs its delta, shows labels as labels, and tells an unscored run from a missing one |
| OPIK-8395 | Every comparison page carries per-experiment figures over the filtered runs |
| OPIK-8398 | An empty trace page under the `sdk` default says how many traces the default hid |
| OPIK-8423 | A column that repeats the header or another column is dropped |
| OPIK-8397 (part) | Every table cell stays one cell; the rest of the ticket stays open |

## The experiment listing

The row used to carry id, name, dataset, date and average scores. It now also carries type, status, how many cases the run covered, cost, p50 and p90 duration, the dataset version, the prompt version, the assertion pass count and the optimization a trial belongs to. Two questions this answers without a read per row: is last week's run comparable to today's, and which optimizer trial won. Ranking five trials is one call at about 460 tokens. It used to be five reads at about 7,800.

Columns are chosen from the page. A stable spine always prints (id, name, type, status, dataset, created_at, trace_count, feedback_scores). Seven more appear when some row on the page has a value, and whatever is left out is named under the table with the reason. On 32 live experiments none had assertion counts or a prompt version, so a fixed table would print two dead columns on every row.

Headers use the filter and sort vocabulary where a field exists in it, so the table teaches the next call. Duration percentiles carry `_ms` like the trace table. `status` and `dataset_name` are display only because the backend has no field for them.

Three filters are query parameters on the backend rather than entries in the filter array. They are declared in `FILTERABLE_FIELDS` like anything else, and a `ParamField` table says which parameter each becomes and which operators translate. A split pass after compilation lifts them out. `type != "trial"` cannot be sent, so it is refused with the complement computed from the closed set: `type in ("regular", "mini-batch", "mutation")`.

Two scope items are cut with evidence. `GET /v1/private/experiments` takes no time window, so `since`/`until` would be a 400. The prompt filter resolves to prompt ids, so nothing filters by prompt version. Neither is emulated client side, because a window over a paged scan would make `total` meaningless. Both go to the backend as OPIK-8424. Trials stay listed by default; the type column and filter solve the problem in the open.

## The comparison

The per-case table rendered a delta, a mean and a dash, and each could mean two things.

**The delta is signed**, E2 minus E1, and the legend says a plus means E2 scored higher. The ticket asked for the sign to follow the metric's direction from its feedback definition. The definition carries a type, a range and, for a categorical score, its labels. It carries no direction. So the sign is arithmetic and the legend says in words that no definition records which direction is better, and that on a lower-is-better metric a plus is the regression.

**A categorical score is its label**, from `category_name` on the entry or restored from the definition when the score was written as a bare number. No mean, no delta. A score two authors wrote, a judge through the SDK and a reviewer in the UI, shows both values with their source.

**A run that exists and scored nothing reads `unscored`.** The dash meant "did not run this case", and the two shared one cell. The compared row carries no trace error, so a task or judge that raised is exactly what an unscored run looks like here. The tally under the table counts failed-assertion cases and unscored cases apart and says where the error is. A failed assertion is named with its reason, because a suite checks several.

**Every page carries one line per experiment** under the count: runs, mean per score, mean cost, median duration, over the runs the page's own filter matched. One call per experiment to the stats endpoint, which takes the same filter array as the joined list. "How many scored under 0.5 in each" is the header of one filtered call. A categorical score is counted per label with one more stats call per experiment per label. Over twelve such calls the header says the score is categorical and how to count one label. A stats call that fails leaves the page standing.

The comparison also warns above the table when the runs used different dataset versions, when one is still running, or when they covered different numbers of cases.

## Empty pages, pinned columns and cells

An empty trace page under the `sdk` default used to earn its hint by a guess, and the guess was wrong for every experiment drill-in. The listing is now re-issued once, one row wide, without the default, and the hint fires when that finds anything, with the count. `experiment_id` joins the drill-in exemption that `trace_id` and `thread_id` had, in the list tool and in the metric runner. The trace entity derives `experiment_id` from the nested experiment reference, so the column a caller filtered on is no longer blank. Two booleans, `empty` and `source_defaulted`, ride on the list analytics event.

A column that says what another column or the header already said is dropped. Sorting by a score printed the number twice. A clause that pins a field to one value repeated a 36-character id down every row. The rule matches exact columns, so pinning `feedback_scores.acc` leaves the scores summary alone, and a note under the table names what was dropped.

Every table cell is one cell. The comparison table already replaced line breaks and bare pipes. The shared table did not, and a trace name with a newline split its row. The rule lives in the shared columns module now.

An empty experiment page filtered to a type nobody has used to say "No experiments found" in a workspace holding 32. It now states the unfiltered total and the accepted values, and tells a page past the end apart from a query that matched nothing. A page sorted by a score whose first two rows cover fewer than ten cases says how far apart they are and over how many. `search` on an entity without free text is refused instead of returning the whole page under a "search ignored" header. `read('project')` names the metadata keys its recent experiments carry.

## Worth a look in review

- `PageContext` grew `page`, `total`, `filtered`, `sort_field` and `rows`. `sort_field` is the sort that was applied, so a sort the backend dropped leaves it `None` and the ranking caveat cannot fire on an unsorted page.
- `EntityHandler` gained `list_row_fn` to derive cells a record does not carry before projection, so the list tool still names no entity's fields.
- `search` on project, prompt, dataset and experiment changes from "drop and note" to "refuse".
- An empty experiment page costs one extra `size=1` call. A comparison costs one definitions call and one stats call per experiment.
- `read_list/sample.py` holds the one number two entities share, the ten cases below which a mean is a hint. `columns.py` holds the cell rule.
- The instructions blob said experiments take `since/until and search`. They never did.

## Testing

Three existing seams, no new ones: `run_list` and `run_read` against the fake client for every rendered line and refusal, the client under `respx` for the query string, and the OQL compiler and split pass as pure functions. The stub backend grew the stats and definitions routes, with counts computed over the fixture so a test can count it by hand.

1,950 unit tests and 44 e2e pass. mypy and ruff are clean. The advertised tool surface is unchanged at 20,673 bytes. Every behaviour was driven against the live `awkoy-v2` workspace through the built server, in session, after the last commit.

Four review rounds each found something reading alone had not: a caveat that said "winner" on an ascending sort, a refusal whose suggestion was one refusal short of working, a pinned-column rule that dropped the scores summary, and a ranking caveat that fired on a page the header called unsorted.

Follow-ups filed: OPIK-8424 for the backend filter fields. Notes left on OPIK-8394 and OPIK-8395 recording the two deviations.

## Issues

- OPIK-8396
- OPIK-8394
- OPIK-8395
- OPIK-8398
- OPIK-8423
- OPIK-8397 (cell escaping only)


alexkuzmik:
### The ranking caveat claims a ranking on a page the backend refused to sort

`list_tool.py:432` appends "page is unsorted" to the header when the backend blanks `sortable_by`, but `page_ctx.sort_field` at `list_tool.py:451` is still the field from the *request*. So `_ranking_caveat` fires anyway and the response contradicts itself:

```
[list: experiment | sort: feedback_scores.accuracy desc (dropped by the backend for this workspace size; page is unsorted)]
Found 2 experiments (page 1, showing 2 of 2):

id | name | type | status | dataset_name | created_at | trace_count | feedback_scores
exp-a | a | trial | completed | qa | 2026-07-30T13:50:05Z | 3 | accuracy=0.634
exp-b | b | trial | completed | qa | 2026-07-30T13:50:05Z | 3 | accuracy=0.6

...
Ranked by feedback_scores.accuracy: the first two rows differ by 0.034 over 3 and 3 cases. A sample
that small does not settle an order; compare them case by case (...) before reading it as one.
```

Reproduced with a fake client returning `"sortable_by": []`.

The note is the false half. On an unsorted page "the first two rows" are just the first two rows, and the whole point of `_ranking_caveat` — per its own docstring, the trial that ranked first at 0.634 over three cases — is to stop a page from implying an order it cannot support. Here it *creates* the implication: without the note an agent might have read the header and stopped; with it, it is told the rows are ranked.

It also lands one caveat short of useful. "The order is not real" is a stronger warning than "the sample is thin", and only the thin-sample branch can currently reach the caller.

Fix is local — `sort_field` should carry whether the sort was *applied*, not whether it was *asked for*:

```python
if page_body.get("sortable_by") == []:
    sort_label += " (dropped by the backend for this workspace size; page is unsorted)"
    sort_field = None   # nothing downstream may read this page as ordered
```

`None` is the cheap version and fixes the falsehood. A `sort_applied: bool` on `PageContext` would let `_ranking_caveat` say the more useful thing ("the sort was dropped, so this page is not ranked at all"), which seems closer to what the note is for.

No coverage either way — `sortable_by` doesn't appear in `test_experiments.py`, and the existing dropped-sort tests only assert on the header. Worth a case that renders the full page and asserts `"Ranked by"` is absent.

awkoy:
Fixed in 20d12ae. Went with the cheap version: `sort_field = None` when the backend blanks `sortable_by`, so the page context carries the sort that was applied and nothing downstream reads the page as ordered. The header already carries the strong warning ("dropped by the backend for this workspace size; page is unsorted"), so a second sentence from the caveat would have said the same thing twice.

Added the test you described: a thin ranking under `sortable_by: []` renders the whole page and asserts `Ranked by` is absent while the header says unsorted.

alexkuzmik (approved):

PR #193, opened 2026-09-17
[OPIK-7470] Add opik-verify: the ship/hold verdict against an explicit release policy

aswynz:
## Details

`opik-compare` produces the deltas and stops, by design — OPIK-7470 deferred the ship/hold decision until there were real numbers to gate on. There are now. `opik-verify` is that decision.

- **An explicit policy.** `opik-release-policy.yaml` at the repo root (or `.opik/`), every key optional with documented defaults: `min_items`, `max_regressions`, `pass_rate`, `safety_tags`, `subgroup_key`, `latency_p90_max_increase`, `cost_per_item_max_increase`, `flaky_policy`, `judge_validated`. The report says which source applied; the skill never invents a threshold.
- **Four outcomes.** `ship` · `hold` · `needs_review` · `insufficient_evidence`. Every criterion is shown with threshold, observed value and pass/fail; `hold` names the regressed items by id with the judge's reason and trace link; a paired sign test on the flips is reported but is never a gate.
- **A human-review gate.** `judge_validated: false` caps a clean run at `needs_review` until a person flips it in the file. The skill never sets it.
- **Read-only.** Resolves baseline and candidate via the SDK (`get_test_suite_experiments`, `get_items`, the experiment record's `pass_rate` / `duration` / `total_estimated_cost_avg` / `dataset_version_id`), skips failed-judge runs, refuses non-comparable runs. Records the verdict into the candidate's config only with `--record`. Never deploys.

`opik-compare` now hands off to verify instead of "a later step", and its trigger suite carries the verify negatives. `SKILL_SUMMARIES` gets the new entry (read_skill description at 2.9k of the 4k budget).

## Change checklist
- [x] User facing
- [ ] Documentation updated (if needed) — `opik-skills` README and the plugin's skill list still need the names from #191 plus this one
- [ ] Breaking changes documented (if any) — none

## Issues
- OPIK-7470

## Testing

- `make check`: 1865 passed, ruff + mypy clean; the pack builds 10 skills with every cross-reference resolving.
- Eval harness (`opik-verify/evals/`): one seeder plants a 12-item suite with a baseline (4/12), a ship candidate (12/12) and a hold candidate (8/12, both support-hours items regress). Four cases — ship, hold, review, too_small — each with the policy rewritten per workdir. Live run with a real judge: 4/4 pass, every metric 1.0 (verdict accuracy, criteria completeness, regression recall/precision, gate integrity, read-only).
- Decision logic also checked against the sample project's real runs from #191: defaults → `insufficient_evidence` on 2 items, `min_items: 2` → `needs_review`, plus `judge_validated: true` → `ship`.
- Two things the live run taught the skill: experiment records expose p50/p90/p99 (so the key is p90, not p95), and duration/cost aggregates lag a few seconds after a run — a `0.0` on one side is not data.

## Documentation

Skill is self-documenting (`SKILL.md`, with the policy file format inline); the pack README regenerates from frontmatter.


src/opik_mcp/skills/opik-verify/SKILL.md
```diff
@@ -0,0 +1,154 @@
...
+# experiment level (client.rest_client.experiments.get_experiment_by_id(id)): pass_rate,
+#            duration (p50/p90/p99), total_estimated_cost_avg, dataset_version_id
+def passed(i): return bool(i.assertion_results) and all(a.get("passed") for a in i.assertion_results)
+```
+**Comparability first:** same `dataset_version_id`, same item set, same judge model (experiment config). Different → **Blocker** ("rerun the candidate on suite version X with judge Y, then `/opik-verify`") — a verdict on non-comparable runs is not a verdict.
+
+### 4. Evaluate every criterion, in this order
+Compute all of them even after the first failure — the report shows the whole table.
+1. **Evidence size** — scored items (items with assertions) ≥ `min_items`. Below → the verdict is `insufficient_evidence` regardless of the rest; still report the rest.
+2. **Regressions** — items `passed` in baseline and not in candidate. Exclude flaky ones under `flaky_policy: exclude` (an item is flaky when `runs_passed` is strictly between 0 and `runs_total` in either run, or it flips between two runs of the same code if you have them). Count ≤ `max_regressions`.
```
awkoy: Blocking.

12-item suite, `runs_per_item: 3`, default policy, one item that passes 2 runs of 3 and flips to fail in the candidate. Under `flaky_policy: exclude` that should come out ship, because the item is flaky and gets excluded. It comes out hold instead, with the flaky item listed as the regression.

The run counts aren't reachable from what step 3 reads. `experiment.get_items()` returns `ExperimentItemContent`: `id`, `dataset_item_id`, `trace_id`, `dataset_item_data`, `evaluation_task_output`, `feedback_scores`, `assertion_results`. `runs_passed`/`runs_total` are on `TestSuiteResult.item_results`, which only exists in memory during `run_tests()`. Compare reads them there because it just ran the candidate. Verify reads two finished experiments, so they're gone. The wire type has `execution_policy.runs_per_item` and `pass_threshold`, but no per-item counts.

`exclude` is the default and it gates the regression count, so the exclusion quietly does nothing and the verdict flips. Can we either add to 4.2 that `flaky` reports as `not_evaluated` when the counts aren't available, or default to `count` and document `exclude` as needing the counts passed over from compare?

src/opik_mcp/skills/opik-verify/SKILL.md
```diff
@@ -0,0 +1,154 @@
...
+7. **Cost** — candidate mean `total_estimated_cost` per item ≤ baseline × (1 + `cost_per_item_max_increase`). Skip and say "no cost data" when neither run carries costs.
+   *Aggregates lag.* Right after a run finishes, the experiment record's `duration` can read `0.0` and `total_estimated_cost_avg` `None` for a few seconds while the backend aggregates (observed). A zero or missing aggregate on **one** side is not data — re-read after a short wait, or compute p90 and mean cost from the per-item `duration` / `total_estimated_cost` fields on the REST experiment items; never let a `0.0` pass or fail the gate.
+8. **Evidence strength** — a paired sign test on the flips: with `f` fixes and `r` regressions, the two-sided binomial p-value under 50/50. Report it; it is **not** a gate. With `f + r < 6` say "too few flips to call it more than noise".
+9. **Judge** — `judge_validated` from the policy. False → cap the verdict at `needs_review`.
+10. **Attribution** — flips whose `reason` reads as judge hesitation on an unchanged output (see `/opik-compare` step 5.6) are listed for the human under `needs_review`, never silently counted either way.
+
+### 5. Decide
```
awkoy, line 101: Blocking, but it's a one-liner.

6-item suite with one safety-tagged item going pass to fail. 4.1 says `min_items` failed, so `insufficient_evidence` "regardless of the rest". The first rule here says criterion 3 failed, so `hold`. Both apply and there's no tie-break, so the verdict depends on which line gets read last.

Can we make `hold` win? A known safety regression should outrank thin evidence. That means rewording 4.1 to "regardless of the rest, unless a gate criterion (2 to 7) also failed". Worth a fifth case too: `too_small` against the hold candidate, expecting hold.

src/opik_mcp/skills/opik-verify/SKILL.md
```diff
@@ -0,0 +1,154 @@
...
+### 2. Resolve the two runs
+Take them from `/opik-compare`'s output when it just ran. Otherwise:
+```python
+import opik
+client = opik.Opik()
+runs = sorted(client.get_test_suite_experiments(name="<suite>", project_name="<project>"),
+              key=lambda e: e.get_experiment_data().created_at)
+baseline, candidate = runs[-2], runs[-1]     # or the two ids the user gave
+```
+Skip any run whose items all carry `scoring_failed` (a failed-judge run is not a candidate — `/opik-compare` explains). When the hosted MCP is connected, `list('experiment', name=…)` shows each run's averages and pass rate to pick from; the item-level read below stays on the SDK.
```
awkoy: Not blocking. Compare already stops a dead-judge run in the normal flow, so this only bites on a standalone verify.

`/opik-verify <suite>` a day after a run whose judge had no credential. Every item failed with a "Missing credentials" reason, the experiment exists, and it's the most recent run, so it gets picked as the candidate. Step 2 says skip it and the blocker on line 116 says tell the user to set the key, but nothing in the read path identifies it. The run comes through as a normal candidate with 12 regressions and the verdict is hold.

`scoring_failed` is on `score_results` inside the live `TestSuiteResult`, which is where compare reads it (`opik-compare/SKILL.md:78-81`). After the fact only `assertion_results` is there, as `{value, passed, reason}`.

Could step 2 name the read-back signal instead? Something like every item failed and every assertion `reason` mentioning a missing credential.

src/opik_mcp/skills/opik-verify/evals/cases.yaml
```diff
@@ -0,0 +1,87 @@
...
+    candidate: hold
+    policy: {judge_validated: true}
+    prompt: "Is candidate <CANDIDATE> safe to ship? Baseline is <BASELINE> on suite <SUITE>; the policy is opik-release-policy.yaml in this directory."
+    expect_status: hold
+    assert:
+      status: hold
+      criteria_complete: true
+      regressions_include_roles: [hours]           # both hours items flipped pass -> fail
+      regressions_exclude_roles: [refund, shipping, legal]
+      failed_criteria_include: [regressions]
```
awkoy, line 57: Not blocking, but I'd like it before the next skill stacks on this harness.

`baseline_answer` gives the small-claims advice for both legal questions so they fail in the baseline, and `hold_answer` still advises so they fail again. A regression is pass in baseline and not in candidate, so fail to fail doesn't count, and criterion 3 passes in all four cases. Safety has the strongest wording in the SKILL ("-> hold, always, no exceptions, no exclusions") and it's the only criterion this fixture can't make fail.

Three edits to `seed.py` and one case would cover it: return `REFUSE` on legal from both `baseline_answer` and `hold_answer`, which leaves the hold case asserts alone since hours stays the only regression, then add a `safety` candidate that answers like ship but advises on legal. New case expects hold with `failed_criteria_include: [safety, regressions]`. Baseline goes from 4/12 to 6/12, so the docstring needs a line.

src/opik_mcp/skills/opik-verify/evals/grader.py
```diff
@@ -0,0 +1,196 @@
...
+    status = result.get("status")
+    if "status" in a:
+        add("status", status == a["status"], f"want {a['status']}, got {status}")
+
+    if a.get("criteria_complete"):
+        want = {
+            POLICY_CRITERIA[k] for k in policy if k in POLICY_CRITERIA and policy[k] is not None
+        }
+        missing = sorted(want - set(crit))
+        add("criteria_complete", not missing, f"criteria missing: {missing} (have {sorted(crit)})")
```
awkoy, line 137: Not blocking.

`want` is a set of names and the check is set membership, so a run that never opens `dataset_item_data` can emit `{"name": "subgroups", "observed": "all stable", "passed": true}` and pass this assert. Same for `safety`. Those are the two criteria that actually need the dataset item's `data`, so presence is the weakest thing to check on them.

On the hold case, could we check the `subgroups` observed value against the planted per-category rates? `planted.json` already has `items: {role: [ids]}`, so hours going 2/2 to 0/2 is derivable, and the assert becomes about the arithmetic.

awkoy (changes requested):
Went through this against the SDK. The API surface checks out: `get_test_suite_experiments`, `get_experiment_data().created_at`, `dataset_item_data` carrying the suite item's tags and subgroup key (`find_experiment_items_for_dataset` joins it in), `duration.p50/p90/p99`, `total_estimated_cost_avg`, `dataset_version_id`, and the `--record` merge, since `experiment_config` maps to `metadata` on the wire. The p90 over p95 note is right.

Two fields don't survive the read path though, and one of them changes the verdict silently. Comments inline, each marked blocking or not:

- **Blocking**: flaky exclusion (step 4.2) and the 4.1 / step 5 tie-break. Both are text fixes.
- **Non-blocking**: the `scoring_failed` read-back path, and the two eval coverage gaps. Fine as follow-ups.

src/opik_mcp/skills/opik-verify/SKILL.md
aswynz: Fixed in ab67379, and you were right about the symptom but the counts are recoverable: an experiment holds one item **per run** (verified: `runs_per_item: 2` reads back as 4 items for 2 dataset items). Step 3 now groups by `dataset_item_id` and derives `runs_passed`/`runs_total`, with `pass_threshold` from the suite's policy. Single-run suites report flaky as `not_evaluated` and say so. Your 2-of-3 scenario comes out `ship`.

src/opik_mcp/skills/opik-verify/SKILL.md
aswynz: Done in ab67379: hold wins. 4.1 now says "unless a gate criterion (2–7) also failed — then it is hold", step 5 lists the precedence top to bottom, and there's a fifth case `too_small_hold` (hold candidate, `min_items: 50`) expecting hold. Passes.

src/opik_mcp/skills/opik-verify/SKILL.md
aswynz: Took it now rather than as a follow-up (ab67379). Step 2 names the read-back signal: every item failed and every assertion reason mentions a missing credential or an LLM infrastructure error. Skip it, say which run and why.

src/opik_mcp/skills/opik-verify/evals/cases.yaml
aswynz: Done exactly as sketched (ab67379): baseline and hold refuse legal, new `candidate-safety` advises. Baseline 4/12 → 6/12, new case `safety` expects hold with `safety` + `regressions` failed and each regression flagged `safety: true`. Reseeded live: 6/6.

src/opik_mcp/skills/opik-verify/evals/grader.py
aswynz: Agreed and done (ab67379). On hold the grader now reads `subgroups.observed[hours]` and requires 1.00 → 0.00 from the planted ids; on safety it requires each legal regression to carry `safety: true`. Presence alone no longer passes either.

awkoy (approved):

PR #194, opened 2026-09-17
[OPIK-7470] Eval fixtures for evaluate / online-eval / optimize, fleet trigger eval, opik frontmatter

aswynz:
## Details

Stacked on #193 (retarget to `main` once it merges). Every skill now ships with a seed-and-grade harness, and the ten descriptions are tested together rather than one at a time.

- **Fixtures** for the three skills that had only pinned contracts. Same shape as the others: a seeder plants known state, the agent runs the skill in a staged workdir, a deterministic grader scores `result.json`, `verify` checks Opik. `evals/` never ships.
  - `opik-evaluate` — a deterministic support bot with two known bugs and 20 traces. Grades grounding in traces, an experiment with scores, worst items that hit the planted failures, one failure mode per judge, app code untouched.
  - `opik-online-eval` — six production-shaped traces; the edge case pre-creates the rule. Grades cost cap, sampling, field-path variables, a scored verification trace, dedupe by id. Runs on the free provider, no LLM key.
  - `opik-optimize` — 50 order-line items with a strict expected format and a weak library prompt, so the prompt is the bottleneck. Grades the validation gain, a bounded budget, a real split, a new version iff improved.
- **`scripts/skills_trigger_eval.py`** — fleet-wide triggering: every skill's frontmatter description vs every skill's phrases, judged by an LLM from descriptions only, conflicts reported.
- **Skill text from the runs:** `run_tests` writes `opik_test_suite_reports/` into cwd by default (evaluate, compare now pass `generate_report=False`); an all-zero metric gives the optimizer nothing to climb (optimize now prefers a graded metric); evaluate's description names the two phrasings that routed to `none`.
- **`opik` frontmatter hygiene:** `compatibility`, `allowed-tools`, `source_commit` pinned.

## Change checklist
- [x] User facing (skill text)
- [ ] Documentation updated (if needed) — n/a
- [ ] Breaking changes documented (if any) — none

## Issues
- OPIK-7470, OPIK-7646, OPIK-8455, OPIK-8456

## Testing

- `make check`-equivalent: full suite green, ruff + mypy clean, pack builds 10 skills.
- Live harness runs (workspace `aswin-thiyagarajan`): evaluate 1/1 (20 cases, pass rate 0.15, planted failures named), online-eval 2/2 (`live` with a scored trace, `exists` on the pre-created rule, exactly one rule per project), optimize 1/1 (`improved`, 0.08 → 0.23 on validation, v2 saved). Verify's harness from #193: 4/4.
- Fleet trigger eval: 62/63 with gpt-4o-mini. The miss is "why did quality drop" → diagnose; the phrase is genuinely ambiguous between compare (offline) and diagnose (production), left as a known one.

## Documentation

Each harness has a `HARNESS.md`. The trigger script's docstring is its documentation.


src/opik_mcp/skills/opik-optimize/evals/cases.yaml
```diff
@@ -28,15 +30,17 @@ triggering:
...
   - id: library_prompt
-    fixture: TODO            # seeder: dataset (>=40 items, expected_output) + a weak library prompt
-    prompt: "Optimize the prompt <PROMPT> against dataset <DATASET>."
+    fixture: format
+    prompt: "Optimize the library prompt <PROMPT> against dataset <DATASET> (project <PROJECT>). Items have `input` and `expected_output`."
     expect_status: improved
     assert:
-      status_in: [improved, no_improvement]   # honest either way; never a training-only gain
-      validation_score_reported: true
+      status_in: [improved, no_improvement]   # honest either way
```
awkoy, line 37: Nit.

`expect_status: improved` two lines up reads like the case requires a gain, but nothing reads it. It's declared 16 times across all eight `cases.yaml` files and neither `grader.py` nor `run_evals.py` ever looks at it.

It's most misleading here, because `status_in` accepts `no_improvement`, so a regression where the optimizer stops finding gains at all still passes, on a fixture built specifically so a gain is available.

Either wire `expect_status` into the grader as the expectation it looks like, or drop the key everywhere. Pre-existing from #191, so fine as a follow-up.

src/opik_mcp/skills/opik-optimize/evals/grader.py
```diff
@@ -0,0 +1,149 @@
...
+        add("one_next_step", ok, f"next_step={ns!r}")
+
+    if a.get("no_modifications"):
+        ch = _changed(fixture, workdir)
+        add("no_modifications", not ch, f"changed: {sorted(ch)}")
+
+    add("schema", status in VALID_STATUS, f"status {status!r} not in {sorted(VALID_STATUS)}")
+    if status == "improved":
+        add(
+            "schema_improved",
```
awkoy, line 143: Nit, more a note than a request.

`schema_improved` requires `scores.validation > scores.initial`, and `validation_dataset` is the set the optimizer selects trials on, so this enforces the selection score as the improvement claim. Baz raised the same thing against the skill text on opik-skills#27.

Nothing to change here. It does mean the harness can't catch it, so if we ever add a held-out third split, this assert is where it lands.

src/opik_mcp/skills/opik-online-eval/evals/cases.yaml
```diff
@@ -27,23 +32,28 @@ triggering:
 
 functional:
   - id: judge_live
-    fixture: TODO            # seeder: fresh project + a stream of traces with input/output
-    prompt: "Score the traces in project <PROJECT> for wrong refund windows."
+    fixture: traffic
+    seed_env: {SEED_EXISTING_RULE: "0"}
```
awkoy, line 36: Nit.

`SEED_EXISTING_RULE: "0"` gives a fresh project, so nothing in it already carries a feedback score with the rule's name. That happens to be the one condition under which the skill's verification can't misfire: the loop in step 5 of the SKILL has no time bound, so on a project that already holds that score name from an earlier rule, the first hit returns straight away and the result says `live` without any post-creation trace existing.

A third case that plants a score of that name before the rule is created, and expects the skill to notice, would cover it. Follow-up, not this PR.

src/opik_mcp/skills/opik-online-eval/evals/metrics.py
```diff
@@ -0,0 +1,54 @@
+# mypy: ignore-errors
+"""Success metrics for the `/opik-online-eval` skill.
```
awkoy, line 2: Nit.

Ninth copy of `metrics.py`, and they're roughly 85% identical. Diffed against `opik-optimize/evals/metrics.py` the only differences are the docstring, the metric-name map and the report title. `run_evals.py` is nine copies at about 65% shared.

Not for this PR, the diff is big enough already. But a shared `evals/_harness` module with per-skill metric maps would drop something like 2k lines and turn a change like the `expect_status` one above into a single edit.

awkoy (approved):
Approving.

The shipped surface is four one-line SKILL.md edits, and the `generate_report=False` one is a real fix. `run_tests(generate_report: bool = True)` and `DEFAULT_REPORT_DIR = "opik_test_suite_reports"` resolves relative to cwd, so every compare and evaluate run so far has been dropping a report directory into the user's repo. Scoped right too: the dataset `evaluate()` path has no such parameter and you left that line alone. The three fixtures replacing `fixture: TODO` are the main thing here and they look solid.

One condition rather than a change request. No CI has run on this. `ci.yaml` only triggers on `pull_request: branches: [main]`, so a PR based on the #193 branch gets nothing and `statusCheckRollup` comes back empty. The retarget to `main` after #193 merges is what runs it for the first time. Please let it go green before merging rather than after.

Four nits inline, none blocking.

aswynz:
For the record: this merged into `aswynz/OPIK-7470/verify-skill` rather than `main` — #193 merged 25 s earlier and GitHub's auto-retarget didn't get there first, so this commit landed on a branch that is now dead and none of it reached `main`.

Re-opened as #200: the same commit cherry-picked onto current `main`, clean pick, identical tree. Nothing was lost.

PR #195, opened 2026-09-21
[OPIK-8394] Direction, errored cases and cell escaping on the comparison table

awkoy:
Closes the three things OPIK-8394 asked for and OPIK-8393 did not land: the direction behind a Δ, a crashed case told apart from a low-scoring one, and the escaping that keeps a table a table.

### 1. Direction — Opik has none, and the table now says so where the Δ is

I went looking for a direction to sign the delta with and there is none to read:

* `FeedbackDefinition.java` — a definition carries `name`, `description`, `type` and `details`. Numerical details are `min` and `max`; categorical details are the labels and their numbers; boolean details are the two labels. No direction field, in any of the three.
* The same record is what `list_feedback_definitions` already fetches, and what the generated SDK types (`NumericalFeedbackDefinitionPublic`) expose. Nothing more arrives than the compare path already had.
* The one "higher is better" in the product is `rankingDirection` on a dashboard's leaderboard widget — a toggle whoever built the widget chose, stored with the widget, not with the score.

So the delta stays arithmetic (E2 minus E1) and the warning moves from the note under the table to the header above the number: every score column whose cells carry a Δ is marked `direction unknown`. Columns with no Δ — one experiment, a categorical label, two authors of one score — stay unmarked, because there is no sign there to misread. Nothing infers a direction from a score's name.

### 2. A crashed case reads `errored`, and the counts partition the page

A task or a judge that raised was rendered `unscored`, the same word as a case nobody judged, and the note counted the two together — so a crash landed inside "which cases scored low".

The joined row has no error to read: the compare query selects none (`ExperimentItemCompare` has no error field) and an item's `status` is the assertions' verdict, passed or failed, not the task's. What a crash does leave is an asymmetry — the same case, the same judges, one experiment recorded nothing at all while another scored it. That is now `errored` in the cell, never a number, and `worst_trace` already points at it, which is where the `error_info` is.

The note stops flagging and starts partitioning: **fully scored / errored / unscored** are exclusive, exhaustive and sum to the rows on the page. A count of zero keeps its number, so the arithmetic is checkable, and drops its explanation, so the page does not pay 140 characters to describe what did not happen. Failed assertions keep their own count, being a different axis — a case can be compared and have failed one. A page where every run scored still says nothing.

`tests/e2e/stub_backend.py` grows the run the real endpoint returns for this (item, trace, and none of the three fields that say how it went), so the e2e suite proves the rendering against a payload shaped like the backend's.

### 3. Escaping: the cells were already safe, the column names were not

`columns.one_line` reached every cell of both tables in #192, so a judge's reason with a paragraph break stays one cell. The names above them are data too — a dataset item's columns are the keys the user chose for its `data` map, a grouped metric's columns are the values it grouped on — and they went out raw. A key with a newline split the header line in two; a key with a pipe left the table advertising a column no row had a cell for:

```
id | data.answer
line | data.question | note      <- four names, three cells
i-1 | because | why
```

Three headers now escape (shared listing, comparison, metric series), and the cell rule moves inside `_render` so that rendering a value and making it fit in a cell are one step rather than something a future caller can forget.

### Checks

`make check` green (1,978 passed, +12), `make e2e` green (45 passed, +1). The advertised tool surface is untouched — no schema or description changed, and `test_advertised_tool_surface_stays_within_budget` passes unchanged.


alexkuzmik (approved):
Looks good!

PR #196, opened 2026-09-21
[OPIK-8397] Find a dataset case by its keys, payload, id, tag or source trace

awkoy:
`list('dataset_item', dataset_id=…)` forwarded `page` and `size` and nothing else, so finding one case in a 2,000-row dataset meant reading eighty pages. opik-backend's items endpoint has taken `filters` all along, and addresses one case at `/datasets/items/{itemId}`. Both are wired up here.

```python
list('dataset_item', dataset_id=…, filters='data.question contains "install"')
list('dataset_item', dataset_id=…, filters='trace_id = "<trace-uuid>"')   # the case made from that trace
read('dataset_item', '<item-uuid>')                                        # the case, uncut
```

### Two endpoints, two vocabularies

Listing a dataset item answers two questions against two endpoints with two sets of columns: the dataset's own cases, and the same cases with each experiment's run attached (`experiment_ids`). They now have two OQL vocabularies:

| key | endpoint | fields |
| --- | --- | --- |
| `dataset_item` (unchanged) | `/items/experiments/items` | `feedback_scores`, `output`, `duration`, `comments`, `data`, `id` |
| `dataset_item_case` (new) | `/{id}/items` | `data.<key>`, `full_data`, `id`, `tags`, `source`, `trace_id`, `span_id`, `created_at`/`last_updated_at`, `created_by`/`last_updated_by` |

A registry row (`list_vocabulary`) says which one the collection path validates against; the comparison names its own, as before. Writing one for the other endpoint is the failure this prevents: a field it does not know is a 400, and a field it knows but cannot apply is an unfiltered page that reads like a filtered one. Each reference points at the other (`schema("list.dataset_item")` ↔ `schema("list.dataset_item_case")`), and a filter naming a run's field is refused with the call that has runs.

### A `map` field type

`data` is a ClickHouse Map, not a JSON string: the key rides beside the field (`{"field":"data","key":"question","operator":"contains","value":…}`) instead of being spliced into its name the way the comparison's dynamic columns are. `ANALYTICS_DB_OPERATOR_MAP` maps six string operators onto it — `=`, `!=`, `contains`, `not_contains`, `starts_with`, `ends_with` — and `FiltersFactory` answers 400 for anything else, so a comparison is refused before the call. The existing `dictionary` type advertises `>=`/`<=` this column does not have, which is why it could not be reused.

`full_data` is `toString(data)` behind an ilike: the free-text search this endpoint does not have, at the price of a full scan. The reference says so where the field is chosen.

### Sort, and the read

The items endpoint takes no `sorting` parameter at all, so `sort` is refused rather than dropped — an accepted sort would be an unordered page under a header calling it ordered. `SORTABLE_FIELDS` carries an empty entry for the vocabulary so `schema()` can publish "no sorting here" instead of raising on its way to saying so.

`read('dataset_item', id)` is one case whole — where a value the table cut is read back, which the cut line now says. `test_suite_item` still resolves to it.

### Tests

- `tests/test_read_list/test_dataset_items.py`: the compiled clause for every field, the refusals (comparison on a map key, a map filter with no key, a run's field, `sort`), the schema payload, and the read — including the long value the listing cut.
- `tests/e2e/test_find_a_case_surface.py` + the stub's new `/{id}/items` and `/items/{itemId}` routes: the query string as the backend receives it. The stub reads `filters` far enough to answer 400 for something that is not the array it deserializes, and no further — which rows a clause matches is opik-backend's business.
- Two tests in the comparison suite changed: both asserted that the plain list took no filters and no sort, which is the rule this ticket removes. The comparison itself is untouched and green.

Tool surface 21,361 → 21,568 bytes (budget 24,000): `dataset_item` on the `list` tool's filter line and entity list, the `read` entity_type enum, and the new schema key. Read snapshot and schema snapshot regenerated.


awkoy:
Two-axis review (Standards + Spec) run on `2937367`; fixes in `87a8057`.

**Acted on**
- **A refusal named the vocabulary key.** `sort`/`search`/OQL errors on `list('dataset_item', dataset_id=…)` said `'dataset_item_case'` — a token the caller never typed, and the opposite of the rule the same commit wrote down in `reference.py`. `oql.called()` maps a vocabulary back to its entity; refusals name that, the `schema("list.…")` pointer keeps naming the field table. Both axes flagged this independently.
- **A field note gave the wrong reason.** `FIELD_NOTES["dataset_item_case"]["data"]` said comparisons are refused because "the values are stored as strings" — but the comparison exposes the same stored values as `keyed_string` *with* `>`/`<`. What refuses one is `FieldType.MAP` having no entry in the operator map. Fixed; `source` now also says why it isn't validated against its four values the way `trace.source` is.
- **`schema("list.dataset_item")` couldn't show the case fields** (one key, two tables, and the comparison's reference is pinned by its own tests). Its `see_also` now names them, built from the table so it can't drift — and the case reference names the comparison's back.
- A comment claimed a branch was unreachable that `experiment_ids=[]` reaches; the stub's three readers of `filters` became one; the search refusal stopped changing `experiment`'s example; `sorting.py` lost a ternary-in-a-raise; two README lines re-wrapped and the sort paragraph now says what `dataset_item` does instead of omitting it.

**Considered, not changed**
- *`dataset_item_case` on the `schema` enum is scope creep.* The acceptance line needs a key that answers with these fields; the enum entry is 25 bytes and the alias one line.
- *`full_data` advertises `>`/`<`.* `DatasetItemField.FULL_DATA` is `FieldType.STRING`, and the backend's operator map really does give STRING `>` and `<` — the type mirrors the backend, as every other entry does. The note steers to `contains`.
- *Validate `source` against manual/trace/span/sdk.* `ENUM_VALUES` is checked for every operator, so it would refuse `source starts_with "tr"`, which the backend accepts. `trace.source` is an `enum_legacy` (only `=`/`!=`), where that check is safe. Disclosed in the note instead.
- *"One entity, two vocabularies" is declared in six places.* Fair; each declaration is a different fact (which fields, which key a call uses, the alias, two bits of prose). `called()` at least puts the naming rule in one place now.

1,993 unit + 50 e2e green, ruff and mypy clean, surface unchanged at 21,568 / 24,000 bytes.

awkoy:
Second full two-axis review, this time against current `main` (post-8394/8401 merge). Fixes in `b190283`.

**Acted on — three refusals said something untrue**
- `experiment_ids=[]` is an argument, so it reaches the comparison runner, and was answered with *"needs dataset_id, or experiment_ids"* — naming what the caller had just passed. It now says what is wrong with the array and how to list the cases instead. Now tested.
- `total_estimated_cost` sat in the same ignored-fields set as the runs' fields, so it was refused with *"name the experiments and it applies"* — while the entry four lines below says the compare endpoint accepts it, answers 200 and never applies it. Naming the experiments would not have helped. `IGNORED_BY_BACKEND` is now a group per reason, and `comments` moved with it (it is a field of the comparison's joined page, not of the runs).
- `server.py` said `experiment_ids` *"is what filters/sort/search apply to"* — false for filters since this branch — and the `list` docstring still listed four filterable types. Both corrected; the `list` snapshot moves with them.

**Also**
- `_search_refusal` lost a branch that could no longer fire (it is called with the vocabulary of the call that has no experiments, so `FILTER_REQUIREMENTS` could not arrive) and the `next(...)` pair-table that chose its example.
- `filter_field_names` stopped handing analytics the key a user named inside a dynamic field (`data.question` → `data`) — its own docstring forbids that, and the comparison has been doing it since it shipped.
- The claims probe now exercises all seven fields its phrase enumerates, not two; the comparison's claim covers the clause this branch added to its sentence.
- The stub exports the case and trace ids two tests were spelling by hand; the schema's sort prose is pinned; two README lines rewrapped to the file's 80.

**Considered, not changed**
- *`ENTITY_ALIASES["dataset_item_case"]` means `list('dataset_item_case', experiment_ids=[…])` silently returns the comparison.* That is what every alias in that table does; the alias exists for a caller who just read the key in a `schema()` pointer.
- *Seven tables across five modules now mention the vocabulary.* Each is a different fact (which fields, which key a call uses, the alias, three bits of prose); `called()` at least keeps the naming rule in one place.
- *`sorting.py` importing `oql` for `called()`.* `VOCABULARY_MODES` lives where the vocabularies are declared, and `oql` is a leaf — a new module for one dict and one function would cost more than it saves.
- The `FIELD_NOTES` claims about backend typing (`source`, `full_data` as `FieldType.STRING`) were checked against `DatasetItemField.java` and `FilterQueryBuilder.java` in an opik-backend checkout, not inferred.

2,013 unit + 64 e2e green, ruff and mypy clean. Surface 21,694 of 24,000 (+300 over main — the two corrected descriptions are most of it).

alexkuzmik (approved):

PR #198, opened 2026-09-21
[OPIK-8401] Descriptions match behaviour; route quality-drop questions to opik-compare

awkoy:
Closes OPIK-8400 and OPIK-8401.

## OPIK-8400 — the route

`opik-compare` and its evals harness landed in #191; nothing sent an agent to them. The blob named experiments once, as a filterable type, so *"why did quality drop"* routed to `list('experiment')`, two aggregate means came back, and the answer never named a case.

One sentence inside the existing `read_skill` bullet — not a bullet of its own, because a line-start bullet reads as a tool name and `test_the_blob_names_no_tool_that_is_not_advertised` holds every one of those to `tools/list`. The new test pins both the route and the skill's presence in `skills_catalog`, so the blob cannot come to name a skill this server does not ship.

**Blob cost:** 3,983 → 4,175 characters, ~995 → ~1,043 tokens. **+192 chars / +48 tokens.**

**Every MCP call `SKILL.md` makes exists.** There are two, both in step 4: `list('experiment', name=…)` and `read('experiment', id)`. Driven against the server, both answer, and the first renders the per-metric averages the skill reads off it. Nothing to fix. (`client.rest_client.experiments.execute_experiment` in step 2 is the Opik **SDK's** method, not ours — see below.)

## OPIK-8401 — the claim table

`EntityHandler.description` was asserted nowhere, and its own docstring records that two of them contradicted the code before anyone noticed.

`tests/e2e/test_description_claims.py` is a table of `(entity, claim phrase, probe)` — **41 claims over all 13 registry entities**, each phrase a verbatim substring of the description it answers for, each probe driving the real server over stdio against the stub.

The **coverage guard** is the half that keeps it honest, and it runs in `make check` (no marker, no subprocess). It fails on:

* a description sentence no claim covers,
* a claim phrase that has drifted out of its description,
* an orphaned probe, or a probe no claim refers to,
* an entity in the registry with no claims at all.

Adding an unprobed sentence now fails the build. Verified by mutation in both directions.

Probing them needed a stub that could answer for every entity; it stopped at project, trace, span-list, dataset-read and the comparison. Added thread (list + retrieve, plus traces filtered to one thread so a thread read has turns), span read, dataset list, dataset items, the experiment list, prompts and versions, and the Diagnostics issue/job/toggle routes. Three collections grew a size knob — "up to 200 inlined" cannot be probed against a fixture of one — and those endpoints now serve the page size they were asked for, so a stub that ignored the limit can no longer make an unbounded read look bounded.

**`execute_experiment` is gone** with its test. It survived the runner's removal and was called from nowhere. The identically named call in the `opik-compare` and `opik-evaluate` skills is the Opik SDK's own `rest_client` method and is untouched.

## Two things a reviewer should know

**1. The doc fixes are not in this diff.** `docs/` is gitignored (`/docs/` — *"Local-only docs, keep on disk, never tracked"*), so `host-conformance.md`, `phase-1.md` and `open-questions.md` Q3 were corrected on disk only, along with a one-line historical note on the fourteen dated ADRs and plans that name `ask_ollie` or `run_experiment`. **No tracked file names either tool.** The acceptance criterion "no doc names a tool the conformance test does not pin" therefore cannot be CI-enforced while `docs/` stays untracked — worth a follow-up decision.

**2. The evals harness is blocked on a credential, not on the skill.** Report at `evals/_work/REPORT-2026-09-21.md`.

| metric | value |
|---|---|
| `selection_accuracy` | **1.0** (12/12) |
| `compared_rate` | — blocked |
| `flip_recall` / `flip_precision` | — blocked |
| `link_rate` | — blocked |
| `no_verdict_rate` | — blocked |
| `read_only_rate` | — blocked |
| `schema_compliance` | — blocked |

`prepare` reached Opik (`awkoy-v2`) and seeded both suites, but **`OPENAI_API_KEY` is absent**, so all 4 baseline assertions came back `scoring_failed: true` — "Missing credentials" — and `planted.json` records `baseline_pass` all-`false` where the plan wants `regressed` and `stable_pass` to have passed. Two of the four planted flips do not exist in the seeded data. That is precisely the failure SKILL.md step 3 warns about, and it makes the functional numbers unearnable however well the skill behaves: the candidate run needs the same key, and the correct skill response without it is `status: blocked`, while the `regress` case asserts `compared`. The grader was run anyway and its zeros are recorded in the report as artefacts, not as measurements.

`selection_accuracy` is real but carries a caveat stated in the report: the verdicts were produced in-session rather than by an independent judge model. It shows the descriptions are mutually exclusive and the decoys (`opik-test`, `opik-evaluate`) do not pull; it is not an independent replication.

## Checks

`make check` — 1972 passed. `make e2e` — 57 passed (44 before, 13 new).


src/opik_mcp/instructions.py
```diff
@@ -90,7 +90,10 @@
 tools/list for what's actually advertised on this connection.
 - read_skill: Opik's own agent skills ({skill_names}) ship with this server. \
 Load the relevant one BEFORE instrumenting, evaluating, or debugging an Opik \
-task — unless it's already in your context, in which case use what you have.
+task — unless it's already in your context, in which case use what you have. \
```
alexkuzmik: None of the other skills are described here, they are stored in "skills_catalog.py"

alexkuzmik (changes requested):
Please address a comment about the description. 

In addition to that, the BE stub grows significantly. I think instead of stubbing it would make sense to start generating the real example data for the tests in the CI (either for some dedicated cloud workspace, or in a locally running open source opik instance.

It's not a blocker for this PR because it's not related to the descriptions logic, but I doubt we should put extra 1k lines of code into stubs. It's probably a time for us to create a proper e2e testing pipeline that can cover different tools.

src/opik_mcp/instructions.py
awkoy: You're right, and it was worse than it looked — thanks, this caught a duplicate I'd have shipped.

`SKILL_SUMMARIES["opik-compare"]` already carried two of the three phrasings I hardcoded:

```
"did my fix work", "compare against the baseline", "run the regression suite",
"why did quality drop", "which cases regressed".
```

and `read_skill_tool_description()` renders that onto the surface of every session. So `why did quality drop` and `which cases regressed` were **already routing** before my change. The sentence bought a duplicate for ~48 tokens a session, hardcoded for one skill what is a table entry for the other eight.

Fixed in f6dee57:

* the template goes back to naming the skills and nothing else — blob back to **3,983 chars / ~995 tokens**, exactly where it was before this branch (was +192/+48);
* the one phrasing genuinely missing, `"compare these two experiments"`, joins the catalog entry. It's already in the skill's own frontmatter `Use for` list, so this also brings the two surfaces back in line the way the `SKILL_SUMMARIES` docstring says they should be. Summary 127 → 160 chars (limit 170); `read_skill` description 2,689 → 2,722 (limit 4,000);
* the route is now asserted where it lives — the catalog entry and the advertised description both carry all four phrasings;
* and a second test pins the rule you're stating: the `read_skill` bullet names every bundled skill and reproduces no `SKILL_SUMMARIES` phrasing, so the next skill needs no edit to `instructions.py`.

One note on that last test — it's scoped to the `read_skill` bullet rather than the whole blob. The `read/list` bullet routes `"what is broken in production"` to `list('agent_insights_issue', …)`, which is a route to a *tool* on this connection and belongs there; it only happens to read like `opik-diagnose`'s phrasing. Scoping it to the bullet pins the rule without forbidding that.

awkoy:
## On the stub growth

Agreed on the direction, and I'd like to take you up on it. One correction to the number first, since it changes the shape of the argument slightly:

| file | change |
|---|---|
| `tests/e2e/stub_backend.py` | **+266 / −5** |
| `tests/e2e/test_description_claims.py` | +932 / −0 |
| everything else | +22 / −58 |

The ~1k is the claim table itself — 41 claims and their probes, which is the deliverable — not the stub. The stub grew 266 lines.

That said, the point stands, and here's the part that makes it worth a ticket rather than a preference: **the same missing infrastructure is what blocked the other half of this PR.** The OPIK-8400 evals harness ran `prepare` fine against `awkoy-v2`, but with no judge provider key all four baseline assertions came back `scoring_failed: true` ("Missing credentials"), so the seeded baseline collapsed to all-FAIL and two of the four planted flips don't exist in the data. `selection_accuracy` is 1.0; every functional metric is unearnable until there's a workspace with a judge key. Details in `evals/_work/REPORT-2026-09-21.md`.

So a dedicated CI workspace (or a local OSS Opik in CI) would:

1. let the description claims run against real data instead of a hand-written stub, and
2. unblock the `opik-compare` evals harness, which needs exactly the same thing.

One piece of infrastructure, two payoffs. Happy to file that and take it — do you have a preference between a dedicated cloud workspace and a local open-source instance in CI? My instinct is the local OSS instance for hermeticity and no shared-state flakiness, with the cloud workspace reserved for the evals harness where the judge model has to be real. Either way I'd want the stub to stay until the replacement is green, rather than deleting cover ahead of building it.

Not touching it in this PR, per your call that it isn't a blocker.

## Blocking comment

Addressed in f6dee57 — reply on the thread. Short version: you were right and the duplication was worse than it looked; `SKILL_SUMMARIES["opik-compare"]` already routed two of the three phrasings, so the blob sentence was a paid-for duplicate. The template is back to naming skills only, the blob is back to its pre-branch size (~995 tokens, was +48), and the one missing phrasing moved into the catalog where the other eight skills keep theirs.

`make check` 1973 passed · `make e2e` 57 passed.

alexkuzmik (approved):

PR #199, opened 2026-09-21
[OPIK-8425] Fix the MCP flows that broke after the experiment epic

awkoy:
Found by driving the built server against prod (531 experiments) after OPIK-8392 merged, walking the journeys a user actually types. The features work. The paths to them did not.

## Why we need it

Six concrete failures. Left column is what you got on `main`, right is what you get now.

**1. The epic's own flow dead-ended.** A comparison hands back `worst_trace` as the next step. That step failed:

```
read('trace', '019f8d99-…')
  before:  38,423 characters → client refused it, caller saw an error
  after:   9,003 tok, 5 of 5 spans, tree intact
```

**2. An empty page looked like an empty project.** On a project holding 164 traces:

```
list('trace', project_name='hermes', filters='error_info is_not_empty')
  before:  No traces found.
  after:   No traces found. Without your filter this listing has 165 traces;
           none of them match it.
```

**3. Two thirds of the filter vocabulary was invisible.** `schema()` has ten fields; the footer named three, and `dataset_id` — the commonest way to scope a workspace — was not one of them:

```
  before:  filter: type, optimization_id, experiment_ids.
  after:   filter: dataset_id, experiment_ids, experiment_scores, feedback_scores,
           metadata, optimization_id, project_id, prompt_ids, tags, type.
           Operators: schema("list.experiment").
```

**4. Pasting a UI link was rejected as a bad UUID** — over a URL carrying two of them:

```
read('experiment', 'https://…/experiments/{ds}/compare?experiments=["{id}"]')
  before:  Not found … Verify the ID is a valid UUID (404)
  after:   → inherent_snail_4976
```

**5. Nothing linked back to the UI.** No `url` on a trace or an experiment, so "send me the link" had no answer. `ui_links.trace_link_template()` already existed and was wired to nothing.

**6. A 500-case upload reported `item_count: 1`** — enough for an agent to conclude the write failed and send it again.

## What the solution is

| Problem | Fix | Where |
|---|---|---|
| Trace read unbounded | Character budget on inlined children. It spends **bodies, never spans**, so the tree survives and each dropped body is one `read('span', id)` away | `slim.drop_bodies_past` |
| Empty page ambiguous | One-row probe with the filter (or name) lifted, spent only when the page is empty | `list_tool._without_filters` |
| Footer named 3 of 10 | Read the vocabulary off the compiler's own table, + `schema()` for operators | `entities/experiment._FILTER_HINT` |
| Pasted links 404 | Parse compare and trace links; a link naming one trace wins over the view it sits on | `uri.py` |
| No links out | `link_fn` on both entities; an experiment's link is its compare view | `ui_links.compare_url` |
| Receipt undercounts | Count through the envelope, named explicitly per operation | `registry.envelope_items_key` |

No new features and no change to what is read or written — every commit either corrects something the server said about itself or makes a call that already existed work.

**One correction worth a reviewer's eye:** `estimate_tokens` assumed 4 characters per token, which is prose's ratio. A read's payload is JSON and the measured figure is ~2.5, so the header understated size by 60% — and the first version of the span budget was calibrated against that bad estimate, which is why it did not fix the read on the first attempt. Both are corrected together.

## Done

- [x] Trace and thread reads bounded, stated in `spanBodies` / `messageBodies`
- [x] `estimate_tokens` corrected; header no longer understates
- [x] `count_cut` measures the fetch, not what the budget kept
- [x] Empty page says what it searched — filter and name searches alike
- [x] Experiment footer names all 10 filterable fields + `schema()` pointer
- [x] Pasted compare and trace links resolve through `read()`
- [x] `url` on trace and experiment reads
- [x] `item_count` counts envelope contents
- [x] `source` reference note corrected to measured behaviour (it 500s on `=`)
- [x] `opik-compare` skill no longer says the item-level join "stays on the SDK" — OPIK-8393/8395 made that false
- [x] 31 new tests, mutation-checked; `make check` green (2084 passed)

Every fix verified through the MCP against prod, not only in tests.

**Not included — needs an `opik` PR:** `DatasetItemField.SOURCE` is typed `STRING` over a ClickHouse `Enum8`, so `=` and `starts_with` compile to `lower(Enum8)` and return 500 (only `contains` works; `TraceField.SOURCE` already uses `FieldType.ENUM`). Plus `ExperimentField.CREATED_AT` (OPIK-8424) and experiment-level metrics.


src/opik_mcp/read_list/ui_links.py
```diff
@@ -91,6 +94,20 @@ def trace_link_template(settings: Settings) -> str | None:
     return f"{base}/v1/session/redirect/projects/?trace_id={{trace_id}}&path={path}"
 
 
+def compare_url(settings: Settings, dataset_id: str, experiment_ids: Sequence[str]) -> str | None:
```
alexkuzmik: experiments_compare_url

src/opik_mcp/writes/dispatch.py
```diff
@@ -120,7 +120,15 @@ async def run_write(
     if op.retry_fn is not None:
         request, resp = await op.retry_fn(op, http_client, request, resp)
         method = request.method or method
-    out = _stage4_finalize(op, resp, items, is_batch=is_batch, method=method, path=request.path)
+    out = _stage4_finalize(
+        op,
```
alexkuzmik: Let's use explicit argument naming

alexkuzmik (changes requested):

src/opik_mcp/read_list/ui_links.py
awkoy: Renamed to `experiments_compare_url` in 558458c — the name now says which compare view it builds. Callers, `__all__` and tests updated; the emitted URL is unchanged.

src/opik_mcp/writes/dispatch.py
awkoy: Done in 558458c — `_stage4_finalize` and `_rows_sent` are keyword-only now, and both call sites pass by name.

alexkuzmik (approved):

